PR #1, opened 2015-07-07
Dev company currency

bago2k4:
card: https://trello.com/c/AHJHZzJX

add a currency picker on top of the page, preselect the currency with the cursor value, pass the currency to all the components that need it.
Where the currency symbol is not present it relies on the currency code (3 chars)

to test:
- `lein clean`
- `lein figwheel`
- load the [page](http://localhost:3449/)
- switch currency
- check that the symbol change real time in finances and compensation

### DIFF
--- a/src/open_company_web/components/compensation.cljs
+++ b/src/open_company_web/components/compensation.cljs
@@ -12,8 +12,8 @@
   [cursor last-values]
   (handle-change cursor (:founders last-values) :founders)
   (handle-change cursor (:executives last-values) :executives)
-  (handle-change cursor (:employee last-values) :employee)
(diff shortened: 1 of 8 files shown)


belucid:
It's a little hard/chaotic to pick your currency from that list. Let's do 2 things.

1st, Make the dropdown list in 2 sections. Common currencies, then a separator, then the rest.

Common will be:

British Pound
Canadian Dollar
Euro
US Dollar

2nd, Make the list of the rest alphabetical.

belucid:
The rest of this is fine, so moving that request to a card and merging this PR.

PR #3, opened 2015-07-08
Dev improve currency picker

bago2k4:
card: https://trello.com/c/EIckWowS

To test:
- Check the currency select
- it should have 4 currencies at the top (CAN, EUR, GBP and USD) and all the other sorted by currency name after a separator
- try to change the currency


belucid:
Add currency symbol and thousand (and million) delimiter to compensation pie chart toolips:

http://stackoverflow.com/questions/13650191/google-chart-number-formatting

PR #8, opened 2015-07-21
Api integration

bago2k4:
card: https://trello.com/c/WLDgwPjX

use the api to load data to show.

- load [localhost:3449](http://localhost:3449), the list of companies should be empty
- add a couple of companies
- reload [localhost:3449](http://localhost:3449), now you should see 2 companies
- click on one, you should only the name of the company
- add a couple of reports for the company
- reload the page, now you should see 2 reports in the list
- click on one, the report should load
- go back and click on the other
- all good? no errors in the console?
- load a company that does not exist: [localhost:3449/companies/BAGO](http://localhost:3449/companies/BAGO)
- load a report that does not exist: [localhost:3449/companies/OPEN/2014/Q8](http://localhost:3449/companies/OPEN/2014/Q8)
- all well? merge

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,10 +12,13 @@
     [org.clojure/clojure "1.7.0"] ; Lisp on the JVM http://clojure.org/documentation
     [org.clojure/clojurescript "0.0-3308"] ; ClojureScript compiler https://github.com/clojure/clojurescript
     [org.clojure/core.async "0.1.346.0-17112a-alpha"] ; Async library https://github.com/clojure/core.async
+    [cljs-http "0.1.35"] ; Http for cljs https://github.com/r0man/cljs-http
     [org.omcljs/om "0.9.0"] ; ClojureScript interface to React https://github.com/omcljs/om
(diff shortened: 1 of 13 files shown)


belucid:
Reviewing.

belucid:
We don't have to use (get ...), when the key is a string can just reverse the order. (:key hash) works and of course ("key" hash) doesn't, but (hash "key") does:

```clojure
({"compensation" 100} "compensation")
100
```

Not a big deal... just keeps the code a little tidier without all the (gets ...)

The other option is when the data comes in from the API you can keywordize it:

```clojure
(clojure.walk/keywordize-keys {"foo" "bar"})
{:foo "bar"}
```

belucid:
Merging this, but let's go back to the keyword data lookup rather than `(get ...)`. Use the keywordize-keys approach on the parsed JSON.

belucid:
@bago2k4 you going to make the updates write back next?

bago2k4:
@belucid yeah i'll change it to use the keywordize-keys tomorrow. i was using `get` because it looked like `(coll "string-key")` was failing if the key was not present but the keywordize approach is better..

PR #11, opened 2015-07-22
Om bootstrap

bago2k4:
card: https://trello.com/c/UsDo12YK

test;
- check the report page
- switch currency
- switch btw the tabs


belucid:
Groovy.

PR #18, opened 2015-08-12
Staging

bago2k4:
review with: https://github.com/belucid/open-company-infrastructure/pull/1

- deploy this branch
- check that api requests are directed to staging-api.opencompany.io
- this should be enough to know that the environment depending vars are ok but if you want you can also introduce an error and check that it appears here https://app.getsentry.com/opencompanyio/oc-staging-web/


src/open_company_web/local_settings.cljs
```diff
@@ -0,0 +1,8 @@
+(ns open-company-web.local-settings)
+
+;; Sentry
+(def local-dsn "https://eab4fe25a2ab4ea1b1d6e10185d63844@app.getsentry.com/48160")
```
belucid: I guess we still haven't been able to move this out?

bago2k4: well we didn't move this out but it will be overwritten when you deploy the repo: https://github.com/belucid/open-company-infrastructure/blob/staging/ansible/web_deploy.yml#L23-L41

PR #20, opened 2015-08-14
Burn rate calc

bago2k4:
card: https://trello.com/c/jWpqiwJC

to test follow the checklist in the card


belucid:
if days of cash remaining is more than 999 we need comma number delimiters

PR #24, opened 2015-09-08
User auth

bago2k4:
card: https://trello.com/c/Ee4lE6Dd

to test follow instructions here: https://github.com/open-company/open-company-auth/pull/1


belucid:
@bago2k4 I assume this needs reworked given all the changes?

bago2k4:
@belucid good too, comments in the other PR

bago2k4:
@belucid this is good to go again

belucid:
Reviewing.

belucid:
On success… we send them back to /login with a JWT arg.

On failure, I modified auth so we send them back to /login with an access arg, and then there are 5 possible reasons why there’s no JWT:

access=**denied**, access=**no-code**, … **invalid-slack-code**, **test-call-error** and **user-info-error**

Those are all the things that can go wrong. The user… really only 2 cases.

They denied it
  -or-
Something is broken

The UI should just distinguish between `access=denied` and `access={anything-else}`.

If it comes back with a denied, we just show them the same Login with Slack: "Add to Slack” button bit and also a message, something like: “OpenCompany requires verification with your Slack team. Please allow access."

With anything else: “There is a temporary error validating with Slack. Please try again later."

If you have `jwt` it’s good and you won’t have `access`.

If you don’t have `jwt`  then check if you have `access`. You might not… cause the user just got there and it wasn’t a redirect, but if you do… you’ll show 1 of the 2 messages.

belucid:
Make login mandatory (wall).

bago2k4:
added a login wall that remember the initial path the user was headed and redirect him there after login.
Also show an error message if login went bad

PR #26, opened 2015-09-14
Cell component

bago2k4:
card: https://trello.com/c/as1uT5T8

#### To test:
- open a report view in editing mode
- scroll at the bottom
- you should see 2 cell: the first is initialized with a fake long value and the second is in the _new_ state
- try to edit both
- try to restore the previous value for both
- try to focus and blur multiple times (sometimes the autofocus for the input field that appears after the cell goes in the _edit_
- if no strange behaviour happens you can merge


belucid:
Really nice!

PR #32, opened 2015-10-27
Redlines

bago2k4:
cards:
- https://trello.com/c/EnmaLrpJ
- https://trello.com/c/0yKU29wv

Incorporate as much as possible of the redlines design we have.
There is still some work to do.

Changes:
- text style
- google charts layout
- collapse section body
- some minor css fixes


belucid:
@bago2k4 not actually pulling in the Proxima nova font.

belucid:
@bago2k4 for some reason it’s 19pt body font when I’m looking at the page at first but if I navigate a section back in time I get 14 pt for the new content

belucid:
@bago2k4 needs a merge of mainline

belucid:
@bago2k4 take a look at Finance notes for Buffer as you move back through time and then come back to the present. It's all dorked up. Presumably because the last revision doesn't have any notes?

bago2k4:
@belucid should be good to go again. I added some cleanup to the code that retrieves and manages the data from the API. It simplified the dispatcher.

PR #33, opened 2015-10-29
Growth section

bago2k4:
card: https://trello.com/c/NCExXzwN

To test:
- use the data imported as the trello card suggests
- change and save the growth and the finances notes
- check all the metrics of all the companies

NB: this also fixes the first 2 elements of the checklist here: https://trello.com/c/NlXTc1vJ

There may be 2 problems:
1) if i PATCH a company with a body of: `{:notes {:body "Body"}}` i get back the section with the same author and updated-at in the section and in the notes
2) importing Thieve Hive i find some metrics that have a target but not a value (check [here](https://github.com/belucid/open-company-infrastructure/blob/master/samples/thrive-hive.edn#L799-L801))


bago2k4:
@belucid this is ready

belucid:
Reviewing.

belucid:
There are a few minor issues, like the link wrapping for Buffer, and showing the link when there is only 1 metric (we should just hide it). But this is good to merge for now.

PR #34, opened 2015-10-29
Design feedback and section sort

bago2k4:
cards:
- https://trello.com/c/NlXTc1vJ
- https://trello.com/c/LaMcOYie

Test:
- check the punch list of the first card
- update a section and check that it's not moved at the top
- refresh the page and check that the previously updated section is still in the same position


belucid:
Reviewing.

PR #36, opened 2015-11-02
Account for images while calculating the section body collapse

bago2k4:
card: https://trello.com/c/7ow2eyi0

to test check on Buffer:
- the _Team and hiring_ section should be collapsed
- also _Diversity_ should be collapsed
- check that all the collapsed sections have the same height


belucid:
@bago2k4 no fade on the diversity image.

bago2k4:
@belucid should be good now

PR #46, opened 2015-11-19
Add sections

bago2k4:
card: https://trello.com/c/bfMlBt1p

to test:
- try to add a section with the link at the top of the ToC
- try to add a section btw other sections and check these facts:
  - the section is added in the right place if the categories match :heavy_check_mark:
  - the section is added at the top if the categories don't match :heavy_check_mark:
  - the category is added if it was empty :heavy_check_mark:
- check CI or run `lein test-all!`
- check the code
- merge

_NB_: saving a new added section is not part of this branch. Also when you select a new section and it gets added to the current list a js error is fired, i was unable to avoid this but the component that errors is unmounted right after so it should not be a problem. ping me for more info if you need

### DIFF
--- a/resources/public/css/style.css
+++ b/resources/public/css/style.css
@@ -136,6 +136,14 @@ div#app {
   margin-top: 50px;
 }
 
+@media (min-width: 960px) {
+  nav.navbar > div {
+    width: 1170px;
+    padding-left: 85px;
+    padding-right: 85px;
+  }
+}
+
 nav.navbar {
   background-color: white;
   border-bottom: 1px solid rgba(0, 0, 0, 0.6);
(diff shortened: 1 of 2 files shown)


belucid:
@bago2k4 please merge mainline in.

belucid:
Reviewing.

belucid:
I don't see the no "Previous" gap fixed on this branch.

belucid:
Section add is being attempted with a PUT, should be PATCH (using the partial-update rel).

bago2k4:
@belucid sorry i merged again the style-fixes branch. But it shouldn't be a problem. About the PUT i didn't do any work on that, all the request you see are old, i was planning on working on the request while working on the section editing branch.

PR #55, opened 2015-12-21
Growth editing

bago2k4:
card: https://trello.com/c/4dZnM5rp

test:
- try to add the growth section for a company that doesn't have it yet
- try to add a new metric to a company that already have the growth section
- try to edit the data and the metadata for an existing metric
- check CI
- merge

NB: there is a problem with cljs-time https://github.com/andrewmcveigh/cljs-time/issues/55


bago2k4:
@belucid tests should be fine now, also i cleaned up some dead code and some style

belucid:
@bago2k4 editing an existing metric (that's not the first one) with the pencil, then clicking save makes the initial metric selected, should stay on the selected metric (like it does when you click cancel).

belucid:
@bago2k4 in the navigation (for finances too) have the selected metric have the color AND bold, but no underline. Have the unselected metrics have the grey and the underline.

belucid:
@bago2k4  Goal dropdown is too fancy. It doesn't need text completion search. What's the "X" to clear it out for?

belucid:
@bago2k4 after adding a new metric it should be selected.

belucid:
@bago2k4 delete button says "DELTE" rather than "DELETE"

belucid:
@bago2k4 from "step 2" I had to press the delete button twice to cancel adding a growth section

belucid:
@bago2k4 from "step 1" pressing cancel button didn't seem to do anything.

belucid:
@bago2k4 typing a metric name wasn't enough to make the next button active, I had to tab out of it.

belucid:
@bago2k4 metric description should be the full width of the component
http://take.ms/qVrwt

belucid:
@bago2k4 Metric name is supposed to be a dropdown combo box, not a simple text field

belucid:
@bago2k4 Interval drop down is too fancy, it doesn't need the type to search

belucid:
Let's simplify the "Measured in". Call it "Measured as" and have it as a simple drop down with just 2 values, "currency" and "numeric". 

In the drop down, "currency" is shown as the company's currency symbol, and "numeric" is shown as "a number".

The default is "numeric"/"a number".

bago2k4:
@belucid all the issues above should be addressed now, also i opened 2 new PR on api and infra to reflect the unit change

belucid:
@bago2k4 Clicking the "More..." link when editing a metric (e.g. MAU in Buffer) gives an error: Uncaught Error: Assert failed: (seq s)

belucid:
@bago2k4 The default "Measured as" for a new metric should be number. Right now it's defaulting to currency.

belucid:
@bago2k4 This is an obscure bug, but:

Start at the company list /companies
Select Buffer
Go to the Key metrics (Growth) section
Click the + to add a metric
Click the back button in your browser
Click the forward button in your browser
You get a metric section w/ no charts

Chrome.

http://take.ms/fniDG

belucid:
Another obscure one @bago2k4 

Add a new metric
Type "Foo" as the name
Backspace and delete "Foo" leaving it blank
Notice, the next button is still active
Click it
Notice the metric has the name "F", even though you deleted the F.

belucid:
@bago2k4 

Select a metric that's not the first one.
Click the + to add a metric
Click the cancel button
Notice the first metric is selected, should stay with the prior selection

belucid:
@bago2k4 after adding a new metric, it's the first one in the list of links. It should be last in the list of links.

belucid:
@bago2k4 the switch between buttons directly under the metric (during step 1 of adding) and buttons under the note (during step 2 of adding) is confusing. Let's leave the note out of step 2 as well.

belucid:
@bago2k4 like when editing a metric measured in currency, the cells should have the currency symbol in addition to the comma/period separators when not being edited:

http://take.ms/lefaf

belucid:
@bago2k4 Once a target gets a value once, seems there is no way to clear it out, make it blank. Even if you blank it, it's 0 when you come back.

belucid:
@bago2k4 Author images are broken, they no longer have a src URL.

belucid:
@bago2k4 

Click + to add a metric to an existing growth section (Buffer)
Click cancel
Notice the metric links and graph are all gone when you're brought back

belucid:
@bago2k4 

when adding a new currency metric, the edited cells didn't have currency symbol:

http://take.ms/dAxiO

bago2k4:
@belucid fixed the 2 issues

PR #58, opened 2016-01-11
Don't force login

bago2k4:
card: https://trello.com/c/xptRIfLA

test:
- remove the JWT token from your cookies
- check some companies data to be sure that everything is working (should be read only since you don't have a token)
- do you see the login button at the top right corner? good (asset is needed for the button)
- click the login button when you are on a company page
- complete the login flow
- do you land back on the same company page? good
- can you edit the company data? (should be if you are part of the same organization)
- merge


belucid:
@bago2k4 please merge in mainline.

belucid:
@bago2k4 

pull

Seems to have issues with Logout. Sequence:

Remove JWT
Visit a company
Login
Click avatar
Click logout button
You end up at the login page... seems good
Return to the company list... you're back to logged in, not good
Return to the company... you're still logged in, not good
Check cookie, it's still there, not good

bago2k4:
@belucid should be addressed now

PR #59, opened 2016-01-12
Company logo

bago2k4:
card: https://trello.com/c/Y62sBxoC

test:
- open a company page
- click on the logo (should bring you to the company profile page)
- insert an invalid url in the logo field (it should fail also if you load a valid url that is not an image: http://asd.com for example)
- unfocus the field, can you see an error message? good
- insert a valid image url in the logo field (https://buffer.com/images/app/buffer-logo@2x.png for example)
- unfocus the field
- does the company logo in the top left corner of the nav bar change? good
- remove the url
- unfocus the field
- does the logo go back to the letter logo? good
- merge


belucid:
@bago2k4 

Updated values for the other items on the page besides logo (name, currency) aren't updated in the PATCH request when you update them.

belucid:
@bago2k4 

If a company logo is provided it should completely replace the default in the top left of the nav. So there should be no round circle and we shouldn't have the company name in text. Just the provided logo.

bago2k4:
@belucid should be good now

belucid:
@bago2k4 please merge in mainline

Also make sure company name isn't in the header if we're using a custom logo.

bago2k4:
@belucid done

belucid:
@bago2k4 I see the company name when a custom logo is used:

http://take.ms/4uaql

bago2k4:
@belucid i did that on purpose, in this case the logo already contains the company name but it won't be always the same, wyt? should i remove the name?

bago2k4:
@belucid company name is shown only when there is no logo

PR #62, opened 2016-01-18
Company not found fix

bago2k4:
card: https://trello.com/c/orcBRl0X

I wasn't able to reproduce the case of the _company not found_ message as it appears to be timing issue with random factors but this should help to avoid that situation.

To test:
- load the companies list
- select a company
- go back
- select another company
- repeat with some more companies
- if you see the _company not found_ message before the company page is rendered please report the exact steps
- if not merge this and hope that it will never happen again :)


belucid:
I still see it.

No real steps. Just fire up figwheel! and visit /

bago2k4:
@belucid should be addressed now

PR #63, opened 2016-01-19
Avoid caching

bago2k4:
card: https://trello.com/c/IXOTgRwZ

Go here for review instructions https://github.com/belucid/open-company-infrastructure/pull/4


belucid:
Reviewing.

belucid:
Seems fine. Would be nice if we didn't have to remember to use the deploy_key stuff in the CSS for images... but, I guess this works.

All images have to come in via CSS, not HTML for this scheme too.

bago2k4:
Well i can write a scss function that given the image name return the image name with the deploy_key attached to it but you'll have to remember to use the function so i guess it's the same..

PR #65, opened 2016-01-23
Devcards fix

bago2k4:
card: https://trello.com/c/EckUGxnz

To test:
- `lein clean && lein devcards!`
- visit [/devcards](http://localhost:3449/devcards)
- check that the devcards work


bago2k4:
@belucid hold on with this, it seems that our test profile is not happy with that exclusion

bago2k4:
@belucid there, i have added a profile for tests so cljs-react-test is included only there, not sure why we didn't do this before but it seems that site, devcards and tests are all good with this configuration.

PR #68, opened 2016-01-23
avoid getDOMNode, use the react component node directly

bago2k4:
card: https://trello.com/c/5ef7Ucz3

this leave out a couple of other warnings, like:
`Warning: React.render is deprecated. Please use ReactDOM.render from require('react-dom') instead.`
and
`Warning: React.unmountComponentAtNode is deprecated. Please use ReactDOM.unmountComponentAtNode from require('react-dom') instead.`
but i can't do nothing about it since they are internally to om.

to test:
- run the local figwheel
- check the site
- the warning were fired simply scrolling the company page down
- if you don't see any warning about getDOMNode deprecated you can merge this


belucid:
Reviewing.

PR #73, opened 2016-02-05
Company Creation w/ placeholder sections

martinklepsch:
Corresponding API PR https://github.com/open-company/open-company-api/pull/30

To test the dashboard:

1. Use `curl` to create two companies. One with your JWToken, one with a different one.
2. Go to both companies' dashboards and see that placeholder sections are only rendered for the company your token created.
3. Use `curl` to `PUT` a "real section" into the company created by the other token.
4. Go to it's dashboard and see that the real section shows up while placeholder sections are still invisible.

To test rest of the flow:

1. Sign out & go to `/` and notice that companies are shown as usual. **confirmed**
3. Sign in & be redirected to a company you have access to or be redirected to `/create-company` if you don't have access to an existing company
3. Create a company on the `/create-company` page. You will be redirected to the newly created company dashboard.


belucid:
@bago2k4 I'm assigning you to this a the primary reviewer since a big chunk of it is refactoring.

I'll QA the new functionality "To test rest of the flow:", but you review the code and regression QA on prior functionality.

belucid:
Reviewing the company create.

martinklepsch:
I'm still "making stuff pretty" so there's still a bit more to come.

belucid:
@martinklepsch we typically would link to the "peer" PR required for this to function in the PR comments: https://github.com/open-company/open-company-api/pull/30

belucid:
@martinklepsch please add instructions for using Boot in the README.

belucid:
@martinklepsch Not required for this PR, but if we keep boot around long term, probably would be good to have it pull its dependencies from one of the lein profiles (dev presumably) rather than duplicate them. If such functionality doesn't already exist for boot, then we should write it, shouldn't be hard.

belucid:
@bago2k4 we'll wait for an all clear from @martinklepsch to review this.

bago2k4:
@belucid ok, i'm waiting

src/open_company_web/actions.cljs
```diff
@@ -0,0 +1,91 @@
...
+(defmethod action :company-created [db [_ body]]
+  (if (:links body)
+    (let [updated (utils/fix-sections body)
+          link (:href (med/find-first #(= "self" (:rel %)) (:links body)))]
+      (router/nav! (str link "/dashboard"))
+      (assoc db (keyword (:slug updated)) updated))
+    db))
+
+(defmethod action :new-section [db [_ body]]
+  (when body
```
martinklepsch: Because `dispatch-body` in the `api` namespace does fall back to an empty map in case of a unsuccessful response these `when` wraps don't do anything. I kept this behavior for now but probably in the future we'll want to clean this up somehow. (And report Xhr errors properly in the UI.)

martinklepsch:
Currently there's no checking if the user may actually create a company. They're only redirected to `/create-company` if they're allowed to but if they navigates to the page on their own it will render just fine. The requests will fail of course but there's no explicit handling for failing Xhrs yet as far as I could see.

martinklepsch:
Oh and I didn't add a test similar to the ones that are already there (i.e. check if some component renders something). I can do that if wanted but it doesn't seem to test much actual behaviour.

bago2k4:
@martinklepsch could you add the curl commands in the PR description? just to make sure i don't lose hours for a header or something like that

martinklepsch:
@bago2k4 I added a the commands to the api projects README: https://github.com/open-company/open-company-api/blob/5221b597051d3e7b9b2d46a23567f47ba6ab89cc/README.md#usage

But good you mentioned it — the command was slightly out of date! Hope you didn't already waste any time :open_mouth:

bago2k4:
@martinklepsch nope, i started from reviewing the code, thanks!

martinklepsch:
@bago2k4 just added some basic tests as well as discussed in slack. I also changed a test command mentioned in the readme, please check if that's correct :)

bago2k4:
@martinklepsch it is correct now, good catch!

bago2k4:
reviewing

bago2k4:
steps worked, reviewing the code now

bago2k4:
@martinklepsch one thing I noticed using boot is that open_company.js still has figwheel stuff inside and this is the result in the console. i'm not sure if we can remove it only when you start the project with boot instead of lein.
http://cl.ly/0B2l032m2n3G

src/open_company_web/actions.cljs
bago2k4: sure, we need to address also responses where the token is invalid and remove it from the cookies and logout the user. @belucid probably this task should be a card itself

src/open_company_web/actions.cljs
```diff
@@ -0,0 +1,91 @@
...
+;; This is a small generic abstraction to handle "actions".
+;; An `action` is a transformation on the app state.
+;; The return value of an action will be used as the new app-state.
+
+;; The extended multimethod `action` is defined in the dispatcher
+;; namespace to avoid cyclical dependencies between namespaces
+
+(defmethod action :default [db payload]
+  (js/console.warn "No handler defined for" (str (first payload)))
+  (js/console.log "Full event: " (pr-str payload))
```
bago2k4: we have a module that wraps the js console, is shodan: https://github.com/noprompt/shodan
no need to change this for now but good to keep it in mind for future work

src/open_company_web/actions.cljs
```diff
@@ -0,0 +1,91 @@
...
+    (let [fixed-section (utils/fix-section (:body body) (:section body) true)
+          assoc-in-coll [(:slug body) (:section body) (:updated-at fixed-section)]]
+      (swap! cache/revisions assoc-in assoc-in-coll fixed-section)
+      (delay-remove-loading))) ; remove loading key
+  db)
+
+(defmethod action :section [db [_ body]]
+  (if body
+    (let [fixed-section (utils/fix-section (:body body) (:section body))]
+      (delay-remove-loading) ; remove loading key
```
bago2k4: this should be the last action because if the :loading key is removed before changing the app-state it could show a gap in the UI, ie: the response is received, the :loading key is removed but the data are not in place yet so the UI shows a message like "Buffer not found", then the app-state is mutated and the :buffer company is in place and the ui render the company. Maybe we can do something like this, avoid to call `delay-remove-loading` and return this:
```
(dissoc (assoc-in db [(:slug body) (:section body)] fixed-section)) :loading)
```
WYT?

src/open_company_web/components/company_editor.cljs
```diff
@@ -0,0 +1,33 @@
...
+            (dom/div {:class "form-group"}
+              (dom/label {:class "col"} "Tagline")
+              (dom/input {:type "text"
+                          :class "form-control"
+                          :placeholder "e.g. Think different."
+                          :value (-> data :company-editor :description)
+                          :on-change #(dis/dispatch! [:input [:company-editor :description] (.. % -target -value)])}))
+            (dom/button {:class "btn btn-primary"
+                         :on-click #(do (.preventDefault %) (dis/dispatch! [:company-submit]))}
+                        "Create Company")))))))
```
bago2k4: @martinklepsch here you can use also the currency picker component but it's not urgent since this will change once we have design.
https://github.com/open-company/open-company-web/blob/mainline/src/open_company_web/components/ui/currency_picker.cljs

src/open_company_web/core.cljs
```diff
@@ -10,6 +11,7 @@
             [open-company-web.components.login :refer (login)]
             [open-company-web.lib.raven :refer (raven-setup)]
             [open-company-web.lib.utils :as utils]
+            [open-company-web.actions]
```
bago2k4: @martinklepsch why do we need this?

bago2k4:
@martinklepsch  once we address/discuss my comments above i think this is good to be merged. i'm passing this to @belucid for final review

src/open_company_web/actions.cljs
martinklepsch: I think because `delay-remove-loading` has a 100ms delay this should not be a problem? Generally I agree that most transformations should happen inside the handlers, easier to test this way.

src/open_company_web/core.cljs
martinklepsch: The `open-company-web.actions` namespace extends a multimethod defined in `open-company-web.dispatcher` if we don't require it it will not be part of the build and thus all dispatched actions won't have an appropriate handler.

src/open_company_web/components/company_editor.cljs
belucid: @bago2k4 we're trying to keep the initial data input to a minimum so we're assuming $ as the currency at company creation time. They can change it later in company settings if it's a wrong assumption.

martinklepsch:
@bago2k4 re figwheel issues: I believe the issue might be caused by figwheel compiling files into `resources/js/compiled/`. Try deleting that folder and if I'm correct the error should be gone after that.

In contrast to lein/figwheel Boot does not write into your project directories but instead acts solely on a classpath it constructed based on your projects sources. Everything Boot does is effectively a transformation on the "Fileset" (immutable abstraction over the filesystem, "filesystem as a value") which can be synced to some directory if desired.

belucid:
Thanks @bago2k4 !

I'll take a final look this weekend and merge after that.

belucid:
@martinklepsch maybe add a task to boot to clear out `resources/js/compiled` if that's the case?

src/open_company_web/actions.cljs
martinklepsch: Thanks, will use next time :+1:

src/open_company_web/actions.cljs
bago2k4: i thought it 100ms was enough but i have seen it fails on staging so i think we should make this change. code will be cleaner too IMHO

bago2k4: NB: it's not a problem with your PR, it was already a problem i'm just using this PR as an opportunity to handle this problem

martinklepsch:
@belucid a task to do `rm -rf resources/js/compiled/? Like one that runs automatically before you run boot to automatically avoid any conflicts?

belucid:
yes @martinklepsch , that's what I was thinking

martinklepsch:
The `sift` task does not actually delete the files but instead just hides them from our boot process. Less invasive than actual deletion.

belucid:
@martinklepsch I logged out and deleted all my companies. Upon logging in, I'm brought to the empty companies list page. I should be brought into the create company flow.

martinklepsch:
@belucid just pushed a fix. No clue how that slipped through, sorry about that!

PR #74, opened 2016-02-12
Topic list edit

bago2k4:
card: https://trello.com/c/cqXKTI8w

NB: this branch also make an effort in the direction of losing jQuery for good. It is still loaded from index.html for now since some old components are still using it but all the new components used in the dashboard doesn't use it anymore.

To test:
- navigate to the dashboard
- tap the bottom button "Manage topics"
- try to reorder the topics
- active/deactivate some topics
- test both button in the navbar: Cancel and Save

PR #75, opened 2016-02-19
Topic edit

bago2k4:
card: https://trello.com/c/FiGpHNm4

This use tinyMCE as main editor for rich content fields.

To test:
- open the dashboard of a company
- expand a topic
- tap on the topic body: you should see the edit button at the bottom of the viewport (if it has no body the button should appear at the previous step)
- edit the topic: try using the new tinyMCE on mobile too if you like (i didn't test it no android as i don't have my tablet with me but i tested it on the iOS simulator and on my iPhone)
- save (or cancel)
- check that the changes are reflected


belucid:
Reviewing.

belucid:
@bago2k4 once I click "Edit Topic" I just get this error:

http://take.ms/eCWqX

It's still deployed on staging if you want to take a look.

bago2k4:
@belucid should be ok now, i forgot to add tinymce functions to the externs file

belucid:
@bago2k4 Let's hide the "Manage Topic" button when the "Edit Topic" button is shown. When they happen to both be near each other it's weird.

belucid:
@bago2k4 I thought we were trying to get the editor icons into 1 row? I see them as 2 rows of 5.

bago2k4:
@belucid i thought that having only one line was just a fallback for not being able to customize the tinyMCE toolbar in two lines, if we like it better i'll try to adjust it with css and then lose the underline and numbered list features if i can't fit all the buttons

bago2k4:
@belucid this PR should be ready but we should wait to merge until #74 is ready too since this include that branch too

PR #76, opened 2016-02-19
Loading animation

bago2k4:
card: https://trello.com/c/qGHFSrBS

NB: this was my idea and may not be in sync with our current designer's choices, probably to be discussed. The gif will be obviously replaced with one more in line with OC.

To test:
- load /companies
- navigate the data of some companies
- check that the gif is always shown when data are loading and is removed when not


belucid:
Reviewing.

PR #77, opened 2016-02-29
Redlines round one

bago2k4:
card: https://trello.com/c/N6wrbezw

To test:
- check page layout with Zeplin
- switch category
- expand topics

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,7 @@
     [com.andrewmcveigh/cljs-time "0.4.0"] ; A clj-time inspired date library for clojurescript. https://github.com/andrewmcveigh/cljs-time
     [funcool/cuerdas "0.7.1"] ; String manipulation library for Clojure(Script) https://github.com/funcool/cuerdas
     [cljsjs/react "0.14.3-0"] ; A Javascript library for building user interfaces https://github.com/cljsjs/packages
-    [medley "0.7.2"] ; lightweight library of useful, mostly pure functions that are "missing" from clojure.core
+    [medley "0.7.3"] ; lightweight library of useful, mostly pure functions that are "missing" from clojure.core
   ]
 
   :plugins [
--- a/project.clj
+++ b/project.clj
@@ -44,7 +44,7 @@
     :dev {
       :dependencies [
         [ring "1.4.0"] ; Web framework https://github.com/ring-clojure/ring
-        [compojure "1.4.0"] ; Web routing http://github.com/weavejester/compojure
+        [compojure "1.5.0"] ; Web routing http://github.com/weavejester/compojure
         [shodan "0.4.2"] ; A ClojureScript library providing wrappers for the JavaScript console API. https://github.com/noprompt/shodan
       ]
     }
--- a/resources/public/js/externs.js
+++ b/resources/public/js/externs.js
@@ -62,6 +62,8 @@ tinymce.activeEditor.getContent = function(){};
 // Sortable.js
 var Sortable = function(){};
 Sortable.create = function(){};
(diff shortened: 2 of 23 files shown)


belucid:
Reviewing.

belucid:
@bago there is a gap between the black line highlighting which category is selected and the bottom of the category nav. That gap isn't there in the redlines.

belucid:
@bago make top right ellipse expose a popup menu with "Logout" and "Settings" as the menu option, and "Company Settings" as an additional menu option if they are looking at a company they have r/w access to.

belucid:
@bago2k4 topic icon color is a grey in the redlines, not black.

belucid:
@bago2k4 there is too much whitespace under the ellipse in the unexpanded topic card

belucid:
@bago2k4 I see you styled the Manage Topics like Trello, that's good. Now we have a treatment from Valdis (in InVision). It's similar, just centered and with more white space. No redlines on it, but eyeball it.

belucid:
@bago2k4 DnD of topics in topic list editing is no longer working in big web.

belucid:
@bago2k4 Company logo needs max height and width (see "open-company" slug on beta for an example)

belucid:
@bago2k4 You see the empty category message while topics are loading.

belucid:
@bago2k4 scroll position isn't handled well anymore. To see this, navigate to "Progress" on a long list like "Buffer", then navigate to "Company" or "Financial", you'll see all the content is scrolled out of view.

belucid:
Scrolling down on the topic edit list causes UI to be borked. Save and Cancel don't stay, but the category nav. does. Save and Cancel need to stay too.

belucid:
@bago2k4 The topic expand animation is add as the ellipse fades out then gives up its space. Same oddness when collapsing as the ellipse reclaims its space.

A more desirable animation is for the ellipse to disappear immediately, for the text of the body to appear in its place and then the topic expand down to reveal the rest of the body text.

Similarly on collapse, the body text should fade away from the bottom and when it's entirely gone the ellipse instantly appear in its place.

bago2k4:
@belucid where are you seeing the gap you talk about in the first comment? i mean what device? i don't see it on chrome (not even with device mode on and the user agent and size set to iPhone 6)

belucid:
@bago2k4 tests are failing on these right now.

bago2k4:
@belucid i retried the tests and now they are ok, the problem was that they exceeded the 10 minutes limit, now they passed in 9 minutes. probably we will have to face this problem again soon.

belucid:
@bago2k4 The 10m timeout is only if there is no log output. We can break things up further into smaller test jobs to get more frequent output if need be.

belucid:
@bago2k4 let's remove the bullhorn stakeholder update link top right. We decided we'll move that under the ellipse menu.

belucid:
@bago2k4 We don't adjust the page length when switching categories. This means short categories have lots of grep space below and long categories don't. Ideally we'd adjust the page length with each category switch.

belucid:
@bago2k4 expanded topics should have date (see Text Topic Expanded and Data Topic Expanded)

belucid:
@bago2k4 expanded topics should have previous / next for history navigation (see Text Topic Expanded and Data Topic Expan

belucid:
@bago2k4 let's have the ellipse menu work and have just "Sign up / Sign in"  or "Logout" for now.

bago2k4:
@belucid ready again, probably we need to iterate a little more on the revision navigation (there is no animation for now)

belucid:
Reviewing again.

PR #78, opened 2016-03-07
Big web

bago2k4:
card: https://trello.com/c/gpApBZ9G

NB: some of the style details has yet to be defined as per the [design feedbacks doc](https://docs.google.com/document/d/1c61v8mkPEjEbDDsWoChjnHRjbSHSvoHGmSQbeGObZYg/edit) so try not to be too sharp.
The type of layout is chosen checking the window width, until 970px is considered mobile, above it it's big web, for now this doesn't take into account the browser's user agent and there is no distinction between smartphones and tablets.

To test:
- check the mock for big web
- try to expand/collapse some topics
- the topic expand animation has to be improved yet: i'm waiting for some feedback on the current one
- expand a topic then click on another topic in the same row: there is no animation for this yet, i'll work on it when i have some feedback on the current expand animation and maybe some guidelines
- check that this doesn't break anything on mobile (activate the chrome device mode or check this on your device)

PR #79, opened 2016-03-08
Mobile ro

bago2k4:
card: https://trello.com/c/weIfykPI

**NB: merge after #78 since this branch is forked from big-web branch.**

To test:
- make sure you are logged in with slack
- select a company with write permissions
- resize you browser window to be at most 970px wide
- try to manage the topics, you can't? good
- try to make the edit button appears, you can't? good
- try to edit a topic in any other way you think you could, you can't? good
- merge


belucid:
Reviewing.

belucid:
@bago2k4 this is ready to go, but I won't merge until the big web PR is ready.

Not specifically related to this work (it's in mainline), but since this is the only mobile PR open, I'll mention it here and let's fix it here. The Prev/Next should scroll so the top of the topic is visible after clicking Prev/Next (unless it already is). Easiest way to see this issue in action is to try "Team" topic on Buffer.

bago2k4:
@belucid fixed

belucid:
@bago2k4 The differences in how date/author attribution are being treated for expanded topics between big web/mobile mocks is unintentional. Should be the same "Joel Gascoigne on November 20" on mobile as it is now on big web.

PR #85, opened 2016-03-29
Light box editing

bago2k4:
card: https://trello.com/c/7xsGcMLv

This card also required this https://github.com/cljsjs/packages/pull/490

To test follow the checklist in the card:
- [x] navigate to `/buffer`, start editing, do NOT change anything, change the url in the bar, you should be able to load the new url
- [x] navigate to `/buffer`, start editing, DO change something (title, headline and/or body), change the url in the bar, test both buttons in the dialog for the correct behavior
- [x] go to `/companies`, select buffer, start editing, hit back, you should go back
- [x] go to `/companies`, select buffer, edit some data of the topic (title, headline and/or body), hit back, test the both the dialog's buttons to make sure it works
- [x] go to `/buffer`, change some data of a random topic, hit save, navigate away of the page, does it works? no dialog is shown? good


belucid:
Reviewing.

belucid:
@bago2k4 editing the title leaves it as is but changes the headline.

bago2k4:
@belucid fixed, too many copy and past from old stuff

bago2k4:
@belucid i fixed almost all, the one item i left out is going to require more time since i can't find why the pencil icon is duplicated when you save. i'll keep looking later, all the rest is addressed

bago2k4:
@belucid not sure but removing the script that replace the svg images with their xml content (we use that to change the color of the icons and it doesn't work very well, sometimes the icons get a gray fill color) seems to fix also the problem with the repeated pencil icon.
i'm thinking more and more to remove this script that is causing a lot of problems since it's mutating the dom w/o React knowing about it and use plain svg or png images with the color we need. we could use frames for the icons too

bago2k4:
@belucid i removed the svg script for now, it seems to have fixed also the problem of the double pencil icon.
I'll handle the remaining things about icons with the dedicated card

belucid:
Reviewing again.

belucid:
@bago2k4 Only issue I still see is the headline counter alignment; http://take.ms/DXyOo

I moved that to the collected issues card so we can close this one out.

PR #89, opened 2016-04-05
Sentry source maps

bago2k4:
card: https://trello.com/c/6VjYrmkf

___

Review with this: https://github.com/belucid/open-company-infrastructure/pull/12
___

NB: the infrastructure PR disable http auth under /js and /lib endpoint. this is needed because Sentry needs to access the source maps and all our js files. If we want to keep it close we can ask them from which subnet they will be accessing and whitelist only those ips.

To test:
- deploy this branch on staging (beta has already this running)
- open buffer on staging and try to expand the finances and growth card
- if you want to try the source maps you need to commit a bug in the code
- deploy on staging or beta
- throw the exception
- check Sentry error info (you should see real funciton and file names, not the compiled names)


belucid:
Reviewing

PR #91, opened 2016-04-06
update 404 handling

martinklepsch:


resources/public/js/set-path.js
```diff
@@ -0,0 +1,13 @@
+function getParameterByName(name, url) {
+    if (!url) url = window.location.href;
+    name = name.replace(/[\[\]]/g, "\\$&");
+    var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)", "i"),
+        results = regex.exec(url);
+    if (!results) return null;
+    if (!results[2]) return '';
+    return decodeURIComponent(results[2].replace(/\+/g, " "));
+}
```
martinklepsch: I'll be up front and honest: I copied this from Stackoverflow, I don't think I fully understand the regex, but it works great 🐬

belucid:
Reviewing.

PR #99, opened 2016-04-11
UI checklist: mobile

martinklepsch:
Various mobile UI cleanups

To review see the individual things to check here:
https://trello.com/c/gwJ1Q2u9/372-collected-ui-issues

Safari mobile issues:
https://trello.com/c/B5ygHR7Q

Also read the comments on that card why certain things haven't been implemented

- [x] **missing**: animation for previous/next inside an expanded topic
- [x]  improve visual balance between icons and title

PR #100, opened 2016-04-11
improve expand/collapse logic

martinklepsch:
Based on #99.

To Test: 

1. open on mobile device
2. navigate categories, check that the autoscrolling is sensible
3. expand a topic
4. scroll, check that topic stays expanded
5. expand other topic, check previously expanded topic remains expanded

### DIFF
--- a/src/open_company_web/components/topic.cljs
+++ b/src/open_company_web/components/topic.cljs
@@ -209,7 +209,7 @@
 
 (defn topic-click [data owner options expanded selected-metric]
   (if (utils/is-mobile)
-    (do (dis/dispatch! [:topic/expand (if expanded nil (keyword (:section-name options)))])
+    (do (dis/dispatch! [:topic/toggle-expand (if expanded nil (keyword (:section-name options)))])
         (mobile-topic-animation data owner options expanded)
         (after 50 #(scroll-to-topic-top (om/get-ref owner "topic"))))
     ((:bw-topic-click options) (:section data) selected-metric)))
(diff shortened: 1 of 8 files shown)


src/open_company_web/components/topic.cljs
```diff
@@ -61,64 +64,61 @@
...
         (let [section (keyword (:section-name options))
               section-data (:section-data data)
               body (utils/get-topic-body section-data section)]
-          (when (clojure.string/blank? body)
+          (when (and (clojure.string/blank? body) (:force-edit-cb options))
             ((:force-edit-cb options) true)))
         ;; hide the edit button if necessary
-        ((:force-edit-cb options) false)))))
+        (when-let [f (:force-edit-cb options)]
+          (f false))))))
```
martinklepsch: I can't find any place where `force-edit-cb` is ever set to anything — is this dead code? @bago2k4

bago2k4: @martinklepsch yep, that was used for mobile editing. I probably missed it when i removed it. good catch

bago2k4:
reviewing

src/open_company_web/components/topic.cljs
```diff
@@ -208,7 +208,9 @@
 
 (defn topic-click [data owner options expanded selected-metric]
   (if (utils/is-mobile)
-    (mobile-topic-animation data owner options expanded)
+    (do (dis/dispatch! [:topic/toggle-expand (keyword (:section-name options))])
+        (mobile-topic-animation data owner options expanded)
+        (after 50 #(scroll-to-topic-top (om/get-ref owner "topic"))))
```
bago2k4: do we really need another function for this? can't we call directly `(js/setTimeout #(scroll-to-topic-top (om/get-ref owner "topic")) 50)` @martinklepsch

martinklepsch:
@bago2k4 I think the `after` was nice because it was a bit easier to read and the time came first. I guess using it once doesn't justify it's own function but it could have been used in other places as well. Potentially be moved to utils haha.

PR #101, opened 2016-04-13
Minor ui issues

bago2k4:
card: https://trello.com/c/YbilftjA/378-minor-ui-issues

To test:
- switch btw dashboard and stakeholder update, the upper part of the page should remain almost unchanged

NB: there is a little jump when you switch, that's due to a calculation needed to center the company logo in the company header. For example in buffer it applies a top margin of 25px to the logo container. We can remove this calculation and avoid the jump but the logo won't be centred.
The problem is that we don't know the logo size until it is loaded. We could also avoid to show the page until the logo is loaded but that's not advisable since the logo is a url field so it could be everywhere and it could take a long time to load.

### DIFF
--- a/scss/partials/_category_nav.scss
+++ b/scss/partials/_category_nav.scss
@@ -8,7 +8,6 @@ div.company-dashboard {
     z-index: 2;
     margin: 0px;
     padding: 2px 0px;
-    width: 100%;
     height: 46px;
     -webkit-backface-visibility: hidden;
     text-align: center;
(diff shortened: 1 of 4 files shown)


belucid:
Reviewing.

belucid:
@bago2k4 initial feedback

- [ ] seems SU picked up dashboard's grey BG below the header, should be white

bago2k4:
@belucid i set the grey background on purpose as i thought we didn't have design yet so i made it look more like the dashboard. anyway i removed that and fixed the width of the container. should be good now

PR #102, opened 2016-04-18
380 minor ui issues

bago2k4:
card: https://trello.com/c/R9OzblPg

To test follow the card.

NB: a couple of items in the card are only striked through but not checked, that's because the fix isn't required anymore due to some other change


bago2k4:
@belucid hold on i'm working on the last issue. i'll update here when it's ready

belucid:
Reviewing

belucid:
@bago2k4 this didn't seem to be done "Make “Cancel” move below “Save” rather than beside it when both are present", but everything else looks great. I moved it to the Part 2 card.

PR #103, opened 2016-04-22
Su editing

bago2k4:
cards:
https://trello.com/c/kVIWFMNy
https://trello.com/c/uRK5Uv8l
https://trello.com/c/GiMyMAto (?)

I added a share button on top of the SU editing page, it's already connected to the POST api call and when it succeeds it redirects the user to the updates list (where he should see the newly created/updated entry).


belucid:
-[ ] Intro (and outro maybe) isn't limited to the same width as topics, should be http://take.ms/8eB4S
-[ ] Foo

belucid:
@bago2k4 going to defer this feedback until after new styling work. Some # of these items will no longer apply or will be addressed by that.

PR #104, opened 2016-04-25
App state refactor

bago2k4:
card: https://trello.com/c/cOLghwj3

The first target of this PR is to avoid composed keys in the app-state. Until now the app-state structure was this:
```clojure
{
  :buffer {
    :name "Buffer"
    :slug "buffer"
    ;; ...other data from /companies/buffer...
  }
  :su-list {
    :collection {
      ;; ... data from /companies/buffer/updates...
    }
  }
  :buffer-su { ;; NB: this key is composed and can cause problems
     :buffer-update-march-2016 {
       ;; ... data form /companies/buffer/buffer-update-march-2016...
     }
  }
}
```

with these changes the new app-state will have this structure:
```clojure
{
  :buffer {
     :company-data {
       :name "Buffer"
       :slug "buffer"
       ;; ... other data form /companies/buffer ....
     }
     :su-list {
        :collection {
          ;; .. data from /companies/buffer/updates...
        }
     }
     :buffer-update-march-2016 {
       ;; ... data from /companies/buffer/buffer-update-march-2016 ...
     }
}
```

This is achieved primarily by centralizing the way the app-state is accessed. From now on we should never access the app-state directly but we should always use the proper functions in the dispatcher.
This way will be very simple to change the structure of the app-state if we will have the need in the future since we just need to change it in one place.

Also this PR introduces the new [urls ns](https://github.com/open-company/open-company-web/blob/app-state-refactor/src/open_company_web/urls.cljs) that centralized the UI urls. This way to change the UI url will be very quick and should avoid any problem of broken links.

Another thing that this PR does is removing the warnings that we see during compilation: `open-company-web.core/history` and `open-company-web/route-dispatch!`.

Last but not least this remove a lot of old components that we don't use anymore.

To test:
- [x] start `boot dev` locally and check that there are warning from 3rd party library only
- [x] access `http://localhost:3559/companies/`
- [x] select Buffer
- [x] add a section and edit it
- [x] try to go back while you have unsaved changes, you should see an alert of unsaved changes
- [x] change the order of the sections in the drawer
- [x] click on Stakeholder Update link in the company header
- [x] check the list of the updates
- [x] select an update
- [x] check that the data are loaded w/o problems
- [x] navigate to `http://localhost:3559/buffer/updates/edit`
- [x] edit the update, save it (this part still under review in another PR but it should work)
- [x] check the 404 page: `http://localhost:3559/asdasd`


belucid:
Reviewing.

PR #106, opened 2016-04-27
386 minor ui issues

bago2k4:
Card: https://trello.com/c/teEmu7vL

to review follow the checklists in the trello card


belucid:
@bago2k4 please merge in mainline.

bago2k4:
@belucid if travis is happy this is ready again..

belucid:
Reviewing.

belucid:
- [ ] Nav. to company profile from drop down menu, then use done button. Have to press done button twice. First time you end up at `http://localhost:3559/buffer/profile?`

bago2k4:
@belucid the missing item i think is server side, let me know if this logic sticks: when you create a company with create-company UI you don't send any of the logo fields to the api :logo or :logo-width or :logo-height. when you get the company after the creation you have those fields containing nil but when you PATCH back logo-width and logo-height with null you get an error because they are optional int values so they should contains 0. WYT?

bago2k4:
@belucid well never mind, i fixed it in the UI. if these fields are null i replace them with 0 for :logo-width and :logo-height and an empty string for :logo

PR #107, opened 2016-04-29
The revolution

bago2k4:
card: https://trello.com/c/NrvNhWeu

I reviewed the charts mostly following the Zeplin design _09 dashboard - mobile v2 - expanded card_.
For now this changes only the charts look but they are still visible only in the expanded card, not when they are collapsed.


belucid:
Reviewing.

belucid:
- [ ] circles should be filled w/ white rather than transparent so the graph doesn't peek through
- [ ] at some point we changed the charts so they stayed on the last point you hovered when your cursor left the chart rather than reverting back to the last data point, this was lost and it's going back to the last data point now

bago2k4:
@belucid i can see the currency symbol on buffer finances:
https://www.evernote.com/shard/s29/sh/1c02b682-ec1c-4189-b253-7a577270ce91/fb58d992fa72da4001a9d160f6955c71

PR #110, opened 2016-05-04
Image uploads, highlighting, list formatting icons

martinklepsch:
[trello card](https://trello.com/c/RDvaxcSa/373-topic-editing-image-handling-highlights-edit-bar-icons)

- [x] Edit an existing topic, check that a plus appears left to the line you're on if the cursor is on an empty line
- [x] click on the plus and upload an image. Save the topic. Verify that the image is still there after a reload.
- [x] check that you can highlight (yellow background) chunks of text using the paint-brush button
- [x] check that the symbols for adding ordered/unordered lists are clear

I'm couldn't find anything better than the paint brush icon but maybe you see something that seems fitting here: https://fortawesome.github.io/Font-Awesome/

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -29,7 +29,8 @@
     [cljsjs/react-dom "0.14.7-0"] ; A Javascript library for building user interfaces https://github.com/cljsjs/packages
     [cljsjs/raven "2.1.0-0"] ; Sentry JS https://github.com/cljsjs/packages/tree/master/raven
     [cljsjs/d3 "3.5.16-0"] ; d3 externs https://clojars.org/cljsjs/d3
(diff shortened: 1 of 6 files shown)


belucid:
@martinklepsch I think the highlight icon you selected is the best of what's there from what I can tell. It works for now.

belucid:
Reviewing.

belucid:
@martinklepsch once you press (+) then move away from it the link hides (which is good) but it's still there and an active link http://take.ms/ozPJO

Would be better if it was truly gone.

belucid:
Everything else looks good @martinklepsch 👍

martinklepsch:
@belucid updated

PR #112, opened 2016-05-11
use filestack from cljsjs

martinklepsch:
@belucid PR got merged so we can now use filestack from CLJSJS. 
Weirdly I don't see any CI check on this PR, does that make sense to you for any reason?


belucid:
@martinklepsch there is a CI check on it now (failing due to emojione)

PR #114, opened 2016-05-13
medium editor toolbar changes (color, buttons)

martinklepsch:
screenshot 2016-05-13 12 43 09

To test:
- run locally
- check that underline and highlight buttons are gone
- check that background color of toolbar is `#4E5A6B`


belucid:
Reviewing.

PR #117, opened 2016-05-17
add image as existing url

martinklepsch:
[trello](https://trello.com/c/TTRanxyC/414-add-image-as-existing-url)

To test:

- edit any post
- move cursor to an empty line
- press (+)
- choose "provide URL"
- paste a URL to any image you like
- hit "add

The image you provided should be uploaded with progress reporting and ultimately be added to the post.

In the background this adds a marker element to the post so that we can replace that marker with the uploaded image later.


belucid:
Reviewing.

belucid:
@martinklepsch externs issue

martinklepsch:
@belucid updated cljsjs package with improved externs.

belucid:
@martinklepsch If I provide a URL that's clearly not an image (e.g. http://cnn.com/ ) it does the uploading thing then just gives me a broken image source. Seems we could learn from filestack that there is no valid image there, yeah?

martinklepsch:
@belucid Filestack doesn't expose any way to detect if an image is good or bad as far as I can tell. Their picker has some special stuff for that but the plain `store` methods don't seem to have any kind of validation. Assuming urls that don't have an image extension are not images is also not really helpful I'd say:
```
https://avatars0.githubusercontent.com/u/12163?v=3&s=96
```
If there's an easy way to do it I'm all for it, just didn't see one (yet?).

belucid:
@martinklepsch seems the best way might be to do a HEAD check on content type of the URL, either original from the user or the one from filestack. But we can tackle that later if and when it becomes a problem.

belucid:
@martinklepsch 

uploaded an image from desktop, worked fine, except the + is still there at the top of the image, and when I select it it says I can upload another one, but it’s on top of the other image

PR #119, opened 2016-05-19
emojione completion improvements

martinklepsch:
[trello](https://trello.com/c/e7vlhS1W/405-emoji-autocomplete)

- [x] use cljsjs emojione
- [x] run emojione in tests
- [x] make sure headline is updated in state before saving
- [x] count emojis as single character in headline count
- [x] make completion UI use emoji sprites

To Test:

run locally and verify that the things above are indeed as described.

- The counting has initial off-by-one errors which are resolved as more characters are typed.
- I also fixed an issue with spaces being counted in weird ways


martinklepsch:
@belucid hold off on this, clojars is rejecting the CLJSJS jar because it's too big, will ping again when resolved.

martinklepsch:
@belucid updated, ready to review.

belucid:
@martinklepsch This regex approach doesn't distinguish between off valid short codes and `:any_old_thing_that_will_stay:`. What about using emojione's function to turn short codes into unicode instead of the the regex, then counting (still probably using the regex to eliminate HTML spaces). This way it'll only count as 1 character if the short code is valid.

martinklepsch:
@belucid updated to only count valid shortnames as single characters

PR #121, opened 2016-05-24
use auth-url instead of `full-url`

martinklepsch:
when we want to have auth and bot adding separate we'll need
different URLs for each flow. This PR uses the new `auth-url`
field which replaces `full-url` (same value, more descriptive name).


martinklepsch:
@belucid ready to review, test steps & more in auth PR: https://github.com/open-company/open-company-auth/issues/5

PR #146, opened 2016-06-02
Custom topics

bago2k4:
card: https://trello.com/c/QuPNopAR

To test:
- add a preset topic
- archive a topic
- add all the preset topics and check that the add topic popover is displayed properly when is smaller
- close the popover
- scroll the page to render the popover above the add topic card and the arrow at the bottom
- add a custom topic
- after adding it you should be sent directly to edit that custom topic with the headline focused and the POST button already visible, check that the editor for the body is working

### DIFF
--- a/scss/partials/_add_topic_popover.scss
+++ b/scss/partials/_add_topic_popover.scss
@@ -121,11 +121,21 @@ div.add-topic-popover {
       font-size: 14px;
       line-height: 1.7;
       color: $oc_gray_5;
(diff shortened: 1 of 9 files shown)


belucid:
Reviewing.

belucid:
- [ ] When a topic is added (Preset, Archived or Custom), topic should open directly to edit mode

belucid:
@bago2k4 this ready again?

PR #147, opened 2016-06-02
improve redirect logic on signup

martinklepsch:
[trello card](https://trello.com/c/5h8tjfIT/450-create-company-flow-for-newly-seen-org-id)

- improves the redirect logic when users sign up
   1. /create-company if they don't have a company yet
   2. redirect via cookie if present
   3. redirect to their company if they have exactly one
   4. redirect to /companies if they have multiple
- adds a few button helper classes that replace the custom styling for save/remove buttons
- improves design on /create-company page to be more in line with Mark's work

**To test**

- [x] log out
- [x] sign in with account that has no company
  - [x] do you get to to `/create-company`?
  - [x] is the name field prepopulated?
  - [x] when creating the company, are you redirected to the new dashboard?
- [x] sign out
- [x] sign in with account that has one or more companies
  - [x] do you get back to the page you were on before? great!

You can check this for the case that the redirect cookie doesn't exist as well but this is unlikely at best so maybe fine to not test explicitly (code is straightforward).

<span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_64.png);background-position:72.5% 27.5%;background-size:4100%" title="rocket"></span></span>

PR #150, opened 2016-06-07
Company user profile

bago2k4:
card:
https://trello.com/c/5t4wTi4o

To test:
- [x] import blank data
- [x] open blank profile
- [x] change name, currency and description and save
- [x] change name and logo (use a valid url that doesn't point to an image, ie: _http://google.com_), you shouldn't be able to save the data, an alert should appear
- [x] change name and logo (use a valid image url)
- [x] open a company page
- [x] navigate to the user info page (from the menu)
- [x] click the "BACK TO DASHBOARD" button, you should go back to the company (this happens only if you used push state to navigate to the profile, if you reload the user profile page you should be sent back to the home page)

NB: I suggested to save the company avatar to filestack but since we have some stuff to finish before the end of the week i prefer to work on more urgent things. Added [this](https://trello.com/c/F0wNvGXp) to the Technical Debt Backlog lane of trello


scss/partials/_back_to_dashboard_btn.scss
```diff
@@ -0,0 +1,19 @@
...
+    font-size: 12px;
+    line-height: 1.5;
+    letter-spacing: 1px;
+    color: rgba(78,90,107, 0.5);
+    margin: 0px;
+    padding: 0px;
+    background-color: white;
+    outline: none;
+  }
+}
```
martinklepsch: These styles button styles are reused across the app. I added a few styles in the `improve redirect logic` PR for this type of button that are easier to reuse: https://github.com/open-company/open-company-web/pull/147/files#diff-36a224998480a3e2239228e57a8a1e76R7

Also consider using the basscss classes `mt1` - `mt4` for margin top and the `p1` - `p4` classes for uniform paddings. 

http://www.basscss.com/#basscss-margin
http://www.basscss.com/#basscss-padding

martinklepsch:
> change name, currency and description and save

should saving maybe close the editing view?

Otherwise working as advertised 👍

scss/partials/_company_profile.scss
```diff
@@ -1,55 +1,197 @@
...
+          height: 30px;
+          background-color: $oc_yellow;
+          position: absolute;
+          border: none;
+          outline: none;
+          font-weight: 700;
+          font-size: 12px;
+          line-height: 1.5;
+          letter-spacing: 1px;
+          color: $oc_gray_5;
```
martinklepsch: @bago2k4 this should also be replaced with `btn-solid`

belucid:
@bago2k4 merge mainline

belucid:
@bago2k4 Remove the description field from this UI.

belucid:
@bago2k4 if you refresh the page while looking at the profile, all is not well.

belucid:
@bago2k4 style the slug in grey so it looks like text on the background rather than looking like the other, editable fields. Put the server URL in front of the slug, eg.: localhost:3000/buffer  staging.opencompany.com/buffer opencompany.com/buffer

belucid:
@bago2k4 I believe there is supposed to be some language about the user info coming from Slack and changing it there, yes?

bago2k4:
@belucid ready again if CI is happy

martinklepsch:
> if you refresh the page while looking at the profile, all is not well.
@bago2k4 after reloading `http://localhost:3559/buffer/profile` the data seems to be still missing

martinklepsch:
I updated @belucid's comments which checkboxes and ticked where appropriate. Reloading the company profile still seems to not fetch the required data

belucid:
@bago2k4  

- [ ] please merge mainline again
- [ ] if you refresh the page while looking at the profile, all is not well. http://take.ms/DzM0H
- [ ] Fix footer to the bottom of the page http://take.ms/2BN80
- [ ] Opening of drawer on big web pushes header down, then over http://take.ms/Q6SYP

bago2k4:
@belucid ready again

belucid:
Reviewing

PR #155, opened 2016-06-20
444 feedbacks

bago2k4:
card: https://trello.com/c/kzyD2ctz

PR #156, opened 2016-06-20
Deep linking

bago2k4:
card: https://trello.com/c/PJxhfRFK

#### This branches from 444-feedbacks and should be merged after #155 

To test section and section editing linking:
- [x] go to /buffer
- [x] go to /buffer/update (if update is a non archived section for buffer)
- [x] go to /buffer/update/edit
- [x] go to /read-only/update/edit while you are logged in: you shouldn't be able to edit
- [x] go to /read-only/update/edit while you are not logged in: you should be redirected to /read-only/update
- [x] go to /buffer and open some sections, navigate using the keyboard arrows btw sections, edit some sections: make sure that the url always reflects the current section and the editing mode

To test history navigation:
- [x] go to /buffer/product
- [x] navigate the history and check that the url is updated to reflect the current as-of
- [x] navigate directly to one revision, ie: /buffer/product?as-of=2015-09-18T18:33:46.000Z
- [x] check that the loaded revision is correct and that you can still navigate the history

To test section navigation inside a stakeholder update:
- [x] create a stakehoder update, let's say the slug is may-2016
- [x] go to /buffer/updates/may-2016
- [x] expand a section, did the url change to /buffer/updates/may-2016/section?
- [ ] use the keyboard arrows to navigate btw sections, did the url change every time the section change?
- [x] close the fullscreen topic, did the url go back to /buffer/updates/may-2016?
- [x] navigate directly to /buffer/updates/may-2016/diversity, did it work?
- [ ] merge

This also address an issue when loading a Stakeholder Update using an ObscURL, if the user didn't have permission for the company GET it crashed the UI. Now it gets the company data needed to show the SU view from the stakeholder update response.

To test:
- [x] create a stakeholder update for a non public company
- [x] share via URL
- [x] open the URL, can you see it? good!
- [x] open the URL from an incognito window, can you see the SU? good!
- [ ] -(optional) open the URL using a slack user of a different slack org, can you see the SU? good!

Possible problem: right now if you navigate to an old revision the content-type for the get request is hardcoded using this function https://github.com/open-company/open-company-web/pull/156/files#diff-f8be44d03d8bcf180c404c8cac0ad27bR20
We should probably change the server side to accept any kind of content-type to avoid errors in the future if we change the content-type.


belucid:
@bago2k4 

This looks really great for section navigation. I didn't see anything in the URLs for history though

- [ ] URL isn't updating with history navigation 
- [ ] arrow right and left didn't do anything in a topic w/ history

bago2k4:
@belucid this is ready, read the description for a possible issue

belucid:
@bago2k4 

- [ ] navigating directly to http://localhost:3559/buffer/update/edit gave me an error
- [ ] going to `/read-only/update/edit` while logged in should also redirect me to `/read-only/update`
- [ ] editing sequence 1) start at `http://localhost:3559/buffer/team`, 2) press the pencil for edit 3) press the browser back button, notice you are still in edit mode even though the URL is back to `http://localhost:3559/buffer/team`.

"Possible problem: right now if you navigate to an old revision the content-type for the get request is hardcoded using this function https://github.com/open-company/open-company-web/pull/156/files#diff-f8be44d03d8bcf180c404c8cac0ad27bR20"

Can you point to a line # instead of a diff?

bago2k4:
@belucid i pointed to the line # of the diff to show you how i updated the function that build the content-type. here is the line:
https://github.com/open-company/open-company-web/blob/deep-linking/src/open_company_web/api.cljs#L19-L20

belucid:
@bago2k4 

- [ ] Go to `http://localhost:3559/buffer/profile` directly or via Company Settings entry in the drawer menu. You get a "blank" section rather than the company settings (this URL should probably be `/buffer/settings` while you are at it)
- [ ] Go to a bad URL `http://localhost:3559/buffer/made-up`, you get a "blank" section rather than a 404

bago2k4:
@belucid ready again

bago2k4:
@belucid NB added some test steps

belucid:
@bago2k4 keyboard navigation on big web between stakeholder sections didn't work

belucid:
@bago2k4 I'm merging this, but let's still get the keyboard nav. fixed here

bago2k4:
@belucid fixed in mainline

PR #160, opened 2016-07-05
Auth flow for unpriviliged users

martinklepsch:
[trello card](https://trello.com/c/yMchqZEe/447-non-bot-user-auth-access) — [Auth PR](https://github.com/open-company/open-company-auth/pull/8) — requires #122 

**To Test**

- [x] try authenticating with a team that doesn't allow integrations being added by usual members (clojurians works well ;))
- [x] you should get back to a page with `access=denied` having some copy about an alternative sign-up link

What remains to be tested: when an integration has been added with the extended scopes can you then sign up with these scopes even though you wouldn't be allowed on your own? If not users would have to go through that process described above every time they login.

PR #161, opened 2016-07-07
Improve email input in su-share form

martinklepsch:
**To Test**

- Go to an SU, click email share, 
- enter email addresses
- remove email addresses
- watch validation messages

**Notable**

- the `:to` key will now contain a list of items instead of a string that needs to be parsed on the server. Server needs to be adjusted for that.
- Validation on the to field happens immediately. I'd like to get to something where the first validation is performed on-blur but this requires
  some more groundwork which doesn't seem right for this ticket (worth a read: http://alistapart.com/article/inline-validation-in-web-forms).
  I've already done some of the groundwork last week but this may need more thought: https://gist.github.com/martinklepsch/e1366008c5a478b33c00d324314da4fd


martinklepsch:
Tests are failing because latest Rum and Om sub-1.0 are incompatible. This is resolved in #160 so we should rebase this on mainline once #160 is merged, which should fix the failing tests.

belucid:
@martinklepsch ready for you again, #160 is merged.

belucid:
@martinklepsch 

This is what I see upon opening email share dialog:

http://take.ms/bYWId

It shouldn't immediately have the validation error, and there is that extra spacing around the text box.

This is what I see when typing in emails: http://take.ms/K8O29

Shouldn't show the validation error until there is some indication the user is done (they type a space, a comma, or field loses focus.

Seems to take 2 clicks to delete an email address. First click seems to just give the field focus. Second click actually removes it.

martinklepsch:
Could you also add a screenshot which rules define the odd border? Also
browser would be interesting. I guess chrome?

martinklepsch:
That extra spacing are your scrollbars — not sure how we fixed that last time do you remember?

martinklepsch:
- ~~I updated the validation error so that it only appears once the user has typed something.~~ Ideally this would only be shown once the input is `blur`ed but this requires a slightly more involved approach to forms & validation to not end in a mess. I have [some ideas around that](https://gist.github.com/martinklepsch/e1366008c5a478b33c00d324314da4fd) but nothing "done" yet. **Update:** fixed that as well via https://github.com/open-company/open-company-web/pull/161/commits/eb4b29ead8e45388c8edb48d74801b2ba16c1765 — I don't think that's a great solution and it violates how one would expect on-change being called but we can revert that once I take some time to work out some proper form validation.
- I couldn't replicate the "two clicks to delete" issue you described, tried all kinds of variations (no email, with emails, with invalid etc).
- Not sure what to do about the scrollbars.
- I also made tabbing between regular inputs and the `item-input` work.

belucid:
👍

PR #164, opened 2016-07-08
MyRedBag debug

bago2k4:
This is to prevent this error or at least to have more info when this happens again:
https://app.getsentry.com/opencompany/oc-beta-web/issues/139841993/

The topic component until now was used for the add topic so there was a lot of code that wasn't used in this case and that could cause the problem. This should help avoiding the error in future.

To test:
- [x] load a dashboard
- [x] try to add a default topic
- [x] try to add a custom topic
- [x] try to add an archived topic


belucid:
Reviewing.

PR #168, opened 2016-07-20
Mainline fixes

bago2k4:
### Please merge ASAP since the current mainline is broken

This fixes an issue that is currently present in mainline. The back-to-dashboard component was migrated from om to rum but in some places was still rendered using `om/build` and it was throwing and error.
Also this removes the edit-topic and expanded-topic components that are not used anymore and simplify a bit the company-dashboard component that had some unused callbacks.

To test:
- visit /buffer
- visit /buffer/update
- visit /buffer/update/edit
- try to make some changes to a section but this shouldn't be a problem since this isn't touching the edit stuff


belucid:
Reviewing.

PR #170, opened 2016-07-22
💸  Getting paid

martinklepsch:
**TO TEST**

- deploy payments service using https://github.com/belucid/open-company-infrastructure/pull/21 & https://github.com/belucid/open-company-payment
- Go to `/buffer/settings` (or another settings page you have access to) 
- Click subscribe and go through the payment form that is opened
- Upon successful submission you should be redirected to the settings page now displaying a "Thank you" message
- Reloading the settings page should eventually also show you the subscription status (i.e. "Subscribe now" or "Manage your subscription")
   I think after you cancel it will still say "Manage your subscription" but I guess that's fine for a V1.

**TODO**

- [x] copy is a mess, @belucid or @stulev need to go over that
- [x] we need Recurly plans for our different environments and these need the right "Success Redirect" URL
- [x] styling of the success message is somewhat meh, I can give that another go but didn't want to spend too much time on it initially

**NOTABLE**

- Subscriptions are tied to `org-id` — I think for now this is ok but eventually we'll need a stable identifier for a company. Slug is changeable so we cant use that.


belucid:
Reviewing.

PR #172, opened 2016-07-29
Simplify topic adding component

martinklepsch:
**TO TEST**

- click "add topic" at end of topic list
- check that no currently used topic is in the list
- check that archived custom topics are properly listed as archived 
- verify that adding suggested/custom and previously archived topics works

**NOTABLE**

- increased contrast of "Add Topic" text
- removed border
- no more pushState (was this needed?)


belucid:
Reviewing.

PR #173, opened 2016-08-02
Emojione picker

bago2k4:
Card: https://trello.com/c/w7tfNFwY

@belucid this needs a one time migration of the db data to change the actual emojis to unicode.

To test:
- try to use the picker in:
- FoCE: headline, snippet
- fullscreen edit: headline snippet (no body)
- SU share: email and slack note
- check that the data sent to the API has only unicode chars and not html img tags
- check that the data coming from API have emojis represented in unicode and that they are translated to html img tags when rendered inside React components


belucid:
Repeatable example of an emoji that doesn't survive round tripping through Unicode: pick the very last flag in the flag category.

Could be transformation to unicode that introduces the issue.

Chrome doesn’t seem to like the Unicode: http://take.ms/bO3Qq (19KB)

transformation is this:
```
encodeURI("&#x"+unicode+";")
```

Tried with the suggested String.fromCharCode but i couldn’t figure out how to trasform the string into a valid hex (using parseInt(uncode, 16) gives a wrong number).

belucid:
taking out the skin tones for emoji” and also, we don’t need a tool tip for emoji

belucid:
@bago2k4 

- [ ] hide the skin tones for emoji (they don't work)
- [ ] we don’t need a tool tip for emoji
- [ ] put the emoji picker inside the note box for email and slack share

belucid:
@bago2k4 please merge in mainline

belucid:
@bago2k4 

- [ ] click pencil to edit a topic, get this warning:
```
react.inc.js:18794 Warning: Failed propType: Invalid prop `search` of type `boolean` supplied to `Picker`, expected `string`. Check the render method of `emoji-picker`.
```

- [ ] sequence: click pencil to edit a topic, click the emoji picker, click outside picker to close it, click, then click the emoji picker again, nothing happens, can't open it any more.

belucid:
@bago2k4 
- [ ] Hide emoji picker if picker is clicked while it is open

belucid:
@bago2k4 

- [ ] in the emoji picker, in the symbols category, there is an emoji for "10" that is showing up w/ no image as `:ten:`

belucid:
@bago2k4 

- [ ] after adding emoji through autocompleter (`:boom:`) it still gets patched as short code rather than getting turned into unicode.

belucid:
@bago2k4 

- [ ] check emoji entered into Slack and Email via autocompleter too, make sure it's POSTed as unicode

bago2k4:
@belucid 
> click pencil to edit a topic, get this warning: ...
that warning is a problem of the emojione-picker, it set the _search_ property to be a string but from their doc you need to set it to true if you want the search to be visible. The component is expecting a string because you can set a search value before you render it.
i think we can safely ignore this warning

bago2k4:
@belucid this should be ready again..

belucid:
@bago2k4 

NB: I turned off the search. Not solely to get rid of the warning, but also because the search functionality was pretty bad. Had to know the exact term. Better/less cluttered to just not have it.

- [ ] sequence: focus on body, click the emoji picker, click on headline to dismiss the emoji picker, notice the emoji picker is still enabled, this is good. Instead focus on body, click the emoji picker, click on the body to dismiss the emoji picker, notice the emoji picker is disabled until you focus on headline then back on body.
- [ ] clicking anywhere outside the emoji picker closes it except clicking the emoji picker icon itself. Clicking the emoji picker icon should close it too when it's open.
- [ ] I see the short code in PATCHed content issue more clearly now: If you select the icon from the autocompleter, you get the unicode, if instead you finish out the short code yourself, by typing the entire `:boom:` or `:robot:` or whatever, it renders fine, but then the short code gets patched rather than the unicode. NB: This doesn't happen in Slack and Email notes. Finishing the short code yourself still posts the unicode.
- [ ] In Slack and Email, ending your note with a typed shortcode `:boom:` `:robot:` or whatever, results in in getting POSTed as not unicode and w/o the final `:`. So a note like:

```
:boom: we did it as :robot:
```

Results in boom POSTed as unicode, but robot posted as `:robot`.

belucid:
👍

PR #176, opened 2016-08-06
UI punchlist

bago2k4:
card: https://trello.com/c/XQ1xFaZd

At the moment the responsive user info view is missing, all the rest should be addressed


belucid:
Reviewing.

belucid:
@bago2k4 This looks really good. Just a couple small items.

- [ ] `:body` isn't shown as placeholder after adding a custom topic
- [ ] now clicking a link FoC gets you both expansion and nav. to the link. Click a link FoC and notice it navs to expansion then navs to the link. Notice if you use the back button, it's back to expanded. Need to not have the expansion take place.

bago2k4:
@belucid done

belucid:
@bago2k4 

- [ ] ordered list doesn't indent the same as bulleted list, it should

Big Web:

- [ ] Restrict add custom topic name to 20 chars (use existing char count location)
- [ ] Emoji icon present during add link http://take.ms/yPaRo
- [ ] Having two primary buttons (gold) on the screen at once is confusing during add image link, seems broken. Instead put the text box/add/cancel below the divider line in the place of the char count/cancel/save. Have those hidden until the user clicks add or cancel.
- [ ] Placeholder text for add image link "http://site.com/img.png". Goes away once they start typing.
- [ ] Adding a link in medium editor w/o protocol (http/https) makes it a relative link. If protocol is missing from what they provide, we should add "http://" to it.
- [ ] While Add button in add topics is disabled, keep it as it is (white), but once they type something in the text box and the button becomes disabled, make it gold.
- [ ] `:ten:` still present in emoji picker

Mobile:

- [ ] Remove swipe between topics
- [ ] Open a topic, scroll is broken.. Very difficult to scroll. When you can scroll, it stops midway through attribution instead of going a bit past it.
- [ ] Opened a topic, tried to scroll, then Return to Dashboard didn’t work. Had to tap multiple times.

belucid:
@bago2k4 mobile issues are more serious and take priority over both/big web.

PR #190, opened 2016-09-22
Lightbox

bago2k4:
Card: https://trello.com/c/G0KwK77E


bago2k4:
The remaining item is going to take some more time to be done properly since we need to mount the div of the lightbox as a body child. This requires to move the current selected topic in the app-state and the fullscren-topic component to be mounted as a root component.

belucid:
Reviewing.

belucid:
@bago2k4 

Tooltips get cut off by the screen edge:

http://take.ms/Pxvc8
http://take.ms/y7iCB

If it's hard to fix those, a work around may be opening them to the right when in lightbox.

belucid:
@bago2k4 

I get a horizontal scrollbar when viewing Buffer's growth.

belucid:
Feels weird to me that we don't open a vertically centered dialog on a tall screen. Stuart also doesn't like that we open over the header. Opening over the header seems OK when it's needed (short screen or super long content), but does not feel great when it's not needed like here:

http://take.ms/wQdK0

belucid:
@bago2k4 

Mobile feedback:

- [ ] Put a CLOSE link (where READ MORE would usually be) in the bottom right of Mobile expanded, so they don't need to scroll back up to the top to tap X
- [ ] After tapping X to close a topic, you lose your scroll position in the dashboard and are back at the top, this is annoying
- [ ] Tap READ MORE on Buffer's Product topic and the content isn't all there

bago2k4:
@belucid 
> Feels weird to me that we don't open a vertically centered dialog on a tall screen. Stuart also doesn't like that we open over the header. Opening over the header seems OK when it's needed (short screen or super long content), but does not feel great when it's not needed like here:

Do you want the navbar to be visible then if you get a enough long screen? so the lightbox overlay should be below the navbar but above the dashboard? This will conflict with the problem that we are having with the scroll that jumps when you open the lightbox (to avoid the dashboard scroll while its open), to do that i have to put the front div under the body so i can't have the navbar visible. A solution would be to remove the opacity background and move the topic box below the navbar.

>  After tapping X to close a topic, you lose your scroll position in the dashboard and are back at the top, this is annoying

This is the same problem of the jumps when opening/closing a topic in the lightbox.

bago2k4:
[READ MORE improvement pass (lightbox)](https://trello.com/c/G0KwK77E/637-read-more-improvement-pass-lightbox)

belucid:
@bago2k4 verified this on iOS Safari and staging:

- [ ] Tap READ MORE on Buffer's Product topic and the content isn't all there

Also:

- [ ] CLOSE should only be in bottom right on expanded topic in Mobile, not on desktop lightbox

belucid:
Reviewing again.

PR #195, opened 2016-10-07
Custom in update

bago2k4:
Always show body placeholder when the body is empty, make sure it's all visible, not only the first line.


bago2k4:
[Placeholder issues](https://trello.com/c/IJmEJTNx/687-placeholder-issues)

PR #198, opened 2016-10-17
Handle auth settings call failure

bago2k4:
This automatically retries every 2 seconds to load the auth-settings from the auth server in case it fails.
Also disable all the auth buttons while the settings aren't loaded.

To test:
- stop your local auth server or the staging service
- logout from the web app
- go to /
- try to login/signup
- [x] is the login/signup button disabled? good
- [x] do you see the auth-settings call retry every 2 seconds in the console? good
- merge

### DIFF
--- a/src/open_company_web/actions.cljs
+++ b/src/open_company_web/actions.cljs
@@ -77,9 +77,12 @@
 
 (defmethod dispatcher/action :auth-settings [db [_ body]]
   (if body
+    ; auth settings loaded
     (when (and (utils/in? (:route @router/path) "confirm-invitation")
                  (contains? (:query-params @router/path) :token))
-        (utils/after 100 #(api/confirm-invitation (:token (:query-params @router/path)))))
+      ; call confirm-invitation if needed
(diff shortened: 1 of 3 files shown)


bago2k4:
[Handle auth settings call failure](https://trello.com/c/qWv8U6gC/697-handle-auth-settings-call-failure)

belucid:
Reviewing.

src/open_company_web/actions.cljs
```diff
@@ -77,12 +77,14 @@
 
 (defmethod dispatcher/action :auth-settings [db [_ body]]
   (if body
```
belucid: @bago2k4 does this take into account the HTTP status? What if we get a body, but not a 200?

belucid: nm, I see it does set the body to false if it's not a success response.

bago2k4: yup, if we want to return special error messages we need to change this but we can still assoc the body here and then show an error message in the UI. For now it should work w/ positive and negative responses

PR #200, opened 2016-10-19
Fix stuck tooltips

bago2k4:
Let all the possible click events bubble up in the DOM tree to the window object.
Set a listener on `window` (and all its descendant) click to remove all the `div.tooltip` nodes


bago2k4:
[Fix stuck tooltips](https://trello.com/c/nTsCjsw5/701-fix-stuck-tooltips)

belucid:
Reviewing.

PR #201, opened 2016-10-19
Share flow updates

bago2k4:
NB: slack sharing still send the old link for stakeholder updates that doesn't include the SU date

### DIFF
--- a/scss/partials/_navbar.scss
+++ b/scss/partials/_navbar.scss
@@ -102,6 +102,21 @@ nav.oc-navbar {
             margin-right: 2px;
           }
         }
+
+        ul.dropdown-menu {
+          li{
+            &:hover {
+              background-color: white;
+            }
+            a{
+              color: $oc_gray_5;
+              &:hover {
+                opacity: 0.6;
+                background-color: white;
+              }
+            }
+          }
+        }
       }
 
       div.user-avatar {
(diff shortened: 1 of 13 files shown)


bago2k4:
[Share flow improvements](https://trello.com/c/Dt5wTYSd/641-share-flow-improvements)

belucid:
Reviewing.

belucid:
- [ ] Update styling of dropdown menu: http://take.ms/iA0j0
- [ ] Update styling of email: http://take.ms/Uykn8

bago2k4:
Here are the updates, @stulev and @belucid i have to say that i don't like the confirmation for slack:
 
screen shot 2016-10-20 at 9 55 58 pm

same goes for email since it uses the same style but it has a done button:

screen shot 2016-10-20 at 9 59 09 pm

What about a title on both?

belucid:
@bago2k4 

- [ ] Add "Email sent!" to the front of the Email confirmation dialog copy
- [ ] Add "Slack message shared!" to the front of the Slack confirmation dialog copy.
- [ ] Remove Done from Email confirmation dialog
- [ ] Have the share button always enabled, and if they are missing a required field (To, Subject) then point it out if they click share (like we do in login forms, etc.)

bago2k4:
@belucid done and deployed to staging

PR #208, opened 2016-11-16
Limit topic body

bago2k4:
Card: https://trello.com/c/eYr3K3AF

Set a characters limit for the topic body, if the user exceeds it while typing (or pasting something) simply disable the save button. Show the char counter.


belucid:
Reviewing.

PR #211, opened 2016-11-30
Conditional nav

bago2k4:
card: https://trello.com/c/3BhPUZqU

Show the Dashboard and Updates links only if there is at least one update shared. When the user creates a new update increase the count in the link for `stakeholder-updates` rel of the company so the links appear.


bago2k4:
[Remove dashboard / updates picker from mobile and dashboard when there are no updates](https://trello.com/c/3BhPUZqU/734-remove-dashboard-updates-picker-from-mobile-and-dashboard-when-there-are-no-updates)

PR #216, opened 2016-12-05
Update derivatives

martinklepsch:
This is still blocked by an actual new release of derivatives but that should happen today or tomorrow so this is a reminder to do it :)

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -12,8 +12,7 @@
 
     [rum "0.10.7" :exclusions [cljsjs/react]] ; https://github.com/tonsky/rum
 
-    ;; Don't upgrade to 0.1.1, errors
-    [org.martinklepsch/derivatives "0.1.0"] ; Chains of derived data https://github.com/martinklepsch/derivatives
+    [org.martinklepsch/derivatives "0.2.0"] ; Chains of derived data https://github.com/martinklepsch/derivatives
 
     [cljs-http "0.1.42"] ; HTTP for cljs https://github.com/r0man/cljs-http
     [prismatic/schema "1.1.3"] ; Dependency of om-tools https://github.com/Prismatic/schema
--- a/src/open_company_web/components/ui/popover.cljs
+++ b/src/open_company_web/components/ui/popover.cljs
@@ -66,7 +66,7 @@
                        ; render the popover component
                        (ru/drv-root {:state dis/app-state
                                      :component #(om/component (component))
-                                     :drv-spec (dis/drv-spec dis/app-state)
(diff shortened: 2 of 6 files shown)


bago2k4:
@martinklepsch thanks for this Martin. While you are looking at it there is an annoying console log that print every time a derivates is created. Can you remove it or add a setting to enable/disable it?
Thank you again.

martinklepsch:
@bago2k4 I think these messages should be gone with 0.2.0 (or maybe even 0.1.1) but not 100% sure. I don't have the time to test these changes right now so if you could check if everything seems to work yourself that would be awesome. 

Also probably best to squash merge this PR as the first commit breaks the build and could hinder later `git bisect` situations...

martinklepsch:
By the way in the `drv-spec` you added a few things like this
```
   :su-list            [[:base] (fn [base] (when (router/current-company-slug)
                                              (-> ((keyword (router/current-company-slug)) base)
                                                  :su-list :collection :stakeholder-updates)))]
```
Ideally every input to the deriving functions should be specified as argument and none should come from global state like `(router/current-company-slug)`. The `path` atom that's defined in the router namespace can be used as a derivative just like the db is used [here](https://github.com/open-company/open-company-web/blob/4a8550bc7b05a22832c9741480701b37d0d09987/src/open_company_web/dispatcher.cljs#L16). 

By not calling functions that retrieve values from reference types (atoms) you can make all the deriving functions pure and thus much easier to understand, debug and test.

The [derivatives docs](https://github.com/martinklepsch/derivatives#usage) only show the usage of one source atom but there's no restriction. If you need to create derived values from multiple atoms you can absolutely do that. (I will update the docs to make that a bit clearer in the future.)

bago2k4:
@martinklepsch as you suggested I updated out drv-spec function to accept also the atom that specifies the current path components as a parameter.
I'll assign this to @belucid now to check my commits but take a look you too if you can.

@belucid to test:
- go to /buff/settings/user-management and check the data
- do you see the list of users? good
- can you resend invitation and delete users? good
- type in an email address and press send invite
- did the user appeared in the table? good
- can you see the invitation call in the network tab of dev tools? good
- merge!

Thanks everybody

martinklepsch:
@bago2k4 looks great 👍 

I hope my suggestions made sense. 😄 If you have further questions just let me know.

bago2k4:
@martinklepsch it did make sense yeah. it's much more cleaner and also safe this way. Thanks a lot!

martinklepsch:
Great to hear :) Always a good feeling to be understood haha 😂

PR #218, opened 2016-12-12
Below the line

bago2k4:
Details [here](https://docs.google.com/document/d/1QZuc0NVTAEzE28n4r-F7yIE8ci9MX4TG0IQd88EIu3A/edit) and [here](https://trello.com/c/2fDHfPrF).


bago2k4:
[P3 Punchlist](https://trello.com/c/2fDHfPrF/741-p3-punchlist)

belucid:
@bago2k4 

- [ ] The right thing isn't happening in regards to displaying topic revisions and timestamps. They have a `:created-at` and an `:updated-at`. The list in the topic detail view is supposed to be using `:created-at` for its ordering and time display. The only place `:updated-at` is used is the `:updated-at` from `author` when showing the tooltip. To see this problem in action. Go to `buff` and edit an older revision. Notice it goes to the top of the list after you save the edit (it shouldn't) and notice the displayed date changes to just now (it shouldn't).

belucid:
@bago2k4 

- [ ] Don't remove "Share Topics" link when in topic details view. If it is clicked on from topic details view, navigate to the dashboard and enter select mode. (Make sure prompting for unsaved changes happens).
- [ ] Rather than "SHARE 0 TOPICS", when there are 0 topics selected, make the copy "0 TOPICS SELECTED" (copy stays the same for 1+ topics selected as "SHARE N TOPICS")

bago2k4:
@belucid feedback should be all addressed. Please pay attention while you review to the created-at updated-at issue you found since it required some changes across multiple files.

PR #223, opened 2017-01-06
Return of FoCE

bago2k4:
card: https://trello.com/c/ZNRFIOF7

### DIFF
--- a/scss/partials/_add_topic.scss
+++ b/scss/partials/_add_topic.scss
@@ -11,6 +11,18 @@ div.add-topic {
   box-shadow:  1px 2px 2px 0 rgba(0, 0, 0, 0.2), inset 0 1px 3px 0 #ffffff;
   padding: 20px;
 
+  div.category {
+    span.category-title {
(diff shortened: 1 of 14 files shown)


bago2k4:
[Return of the "in place" FoCE](https://trello.com/c/ZNRFIOF7/762-return-of-the-in-place-foce)

belucid:
Reviewing.

belucid:
@bago2k4 This mostly looks great.

1 big issue. The FoCE edit (as opposed to new entry) is using the `update`/PUT (creates a new revision) rather than `partial-update` which edits.

I'll resolve lack of the link on the API side.

belucid:
@bago2k4 

- [ ] Use Fundraising and Finances in a company, and the header doesn't appear in the list (great!) but it leaves a gap (not good) http://take.ms/4JETi

belucid:
@bago2k4 

- [ ] missing 3rd "point to the little +" tooltip

PR #231, opened 2017-03-21
Access control enhancements and Boards

bago2k4:
Trello cards:
https://trello.com/c/DgZPaV90
https://trello.com/c/DQDPbcNg
https://trello.com/c/ihjBQDRs
https://trello.com/c/htV1KDYO
https://trello.com/c/w6Pp1gnP
https://trello.com/c/3oSOhvfa

This PR is needed to support the access control changes made in Auth and Api services. It introduces the concept of team, organization and board.
Also this move all the old code from `open-company-web` to the new `oc.web` namespace for 2 main reasons:
- respect the namespace pattern we use in all the other services
- move all the needed code and leave out the old not anymore necessary code and components

Also this tries to make another step ahead in the move from Om to Rum since it migrates some components.

PR #232, opened 2017-03-27
Attachments

bago2k4:
___
### NS: Merge after #231 and #230 
___

Card: https://trello.com/c/T22m9Om2
Related PR: `open-company-api/attachments` TBD

To test:
- from dashboard edit a topic
- add a file from local device
- add another from google search or from gdoc
- check the list of attachments, everything looks correct? the attachments says `Draft`? good
- save the topic, do you see the creation date instead of `Draft`? good
- check `open-company-attachments` bucket of our S3, do you see the files you just uploaded? good
- refresh the page
- do you still see the list of files? good
- enter topic view for the topic, do you see the list of files? good
- edit the last entry
- remove a file
- add another
- refresh the page? do you see the expected files? good
- merge
- do one of Sean's funny things at the end of PRs

### NB: review with this https://github.com/belucid/open-company-infrastructure/pull/36


bago2k4:
Once this https://github.com/cljsjs/packages/pull/1069 is merged tests should be ok.. now they fail because it can't find the package on clojars.

belucid:
@bago2k4 Reviewing

belucid:
@bago2k4 Looking in the S3 bucket. How come some are "standard" and others are "reduced redundancy" storage class? Seems almost random. I can't discern the pattern.

belucid:
@bago2k4 Don't love having 1 bucket for all environments' attachments. It's mixing local, staging and beta (prod). Better to have, at a minimum an S3 bucket for non-prod (local, staging) and prod (beta for now) so we can dump the non-prod stuff at will.

belucid:
@bago2k4 list of attachments seem to show in random order (at least I can't discern a pattern). Probably better to show them in chronological order by when they were added, or alphabetical.

belucid:
@bago2k4 click remove attachment w/ the tooltip showing. It gets stuck:

http://take.ms/jmeAE

bago2k4:
@sean about the Reduced Redundancy that was set in the filestack app setup. We have some files w/ this class and some with the STANDARD class because we still have the old app running somewhere and that's uploading files w/o the Reduced Redundancy class.
I removed this flag from our current app too since it doesn't seem desirable for our user content to have less redundancy even if we can save some bucks, we can talk about this more if the storage become a big expense.

bago2k4:
@sean this is ready again. Now localhost and staging are using `open-company-staging-attachments` while beta and production use `open-company-attachments` bucket.
Both use the same CORS and Policy configuration so there shouldn't be problems while uploading.
The issue about the stuck tooltip should be addressed too.
Ready again for review.

bago2k4:
@belucid now this needs this (infra PR)[https://github.com/belucid/open-company-infrastructure/pull/36] to work, review that together with this and use that branch to deploy.

bago2k4:
@belucid i forgot about the tests, looking into it now. i know why they are failing but i don't know the fix just yet.
Using cljsjs package for jwt-decode lib triggers an error on Raven lib when compiling for tests. Weird.

bago2k4:
@belucid tests should be fine now, ready again.

belucid:
Reviewing again.

PR #233, opened 2017-03-31
History count

bago2k4:
___
### Merge after #231 
___

To test:
- import 18F to your team
- visit `/18f/all-hands`
- make sure that only `press` topic has the prior history indicator
- merge


bago2k4:
Once this https://github.com/cljsjs/packages/pull/1069 is merged tests should be ok.. now they fail because it can't find the package on clojars.

bago2k4:
[Topic history indicator on initial board load](https://trello.com/c/knMG0hRu/790-topic-history-indicator-on-initial-board-load)

PR #234, opened 2017-04-10
Assets packages

bago2k4:
Card: https://trello.com/c/GEY2hcA2

### Review with: https://github.com/belucid/open-company-infrastructure/pull/35

To test:
- deploy to staging using the `sync_images=true` setting
- deployed? good
- make sure that staging works and that the following files are loaded from the right CDN:
  - oc.js
  - oc_assets.js
  - main.css
  - all the images except of emojione.sprites.png (that is loaded always from the production CDN)


belucid:
@bago2k4 tests are failing here

belucid:
@bago2k4 it's really not `staging`, it's `non-prod` since we use it for local development too, so I have a slight preference for using that bucket label.

Also, probably less confusing when looking at the buckets in S3 if we put the environment last, so that the buckets are listed together in the S3 UI. So:

`open-company-assets`
`open-company-assets-non-prod`
`open-company-attachments`
`open-company-attachments-non-prod`

Rather than:

`open-company-assets`
`open-company-staging-assets`
`open-company-attachments`
`open-company-staging-attachments`

bago2k4:
@belucid i rolled back wurfl.io from being loaded from our server since the content of that js changes based on the device that is loading the script. We need to load it from it's own source or it will lose it's functionality.

bago2k4:
@belucid this is ready again

PR #239, opened 2017-05-18
Comments

bago2k4:
Part 1 of this card: https://trello.com/c/aBhDvqzr

Static comments. This is missing the delete and edit comments. For now it retrieves the comments for each rendered entry in the topic view. When the user clicks on the comments icon it toggles the comments UI.
The comments button is shown only if there is the `comments` rel in the entry data. The new comment component at the bottom of the comments container is shown only if the is the `comment` rel in the entry area.
So the comments icon won't be shown on a new entry (not yet saved).

To test:
- open a board and click on a topic (with more entries if possible)
- check the number of comments besides the comments icon (shown only if >0)
- click it
- check that the comments are shown
- add a comment
- check that the number of comments is increased in the comments button
- refresh the topic view and check that the number of comments is still good
- toggle the comments for more entries to make sure the right data are always shown

### DIFF
--- a/resources/public/js/externs.js
+++ b/resources/public/js/externs.js
@@ -68,6 +68,7 @@ var OCWebConfigLogLevel = function(){};
 var OCWebPrintUpdatesListData = function(){};
 var OCWebPrintUpdateData = function(){};
 var OCWebPrintRouterPath = function(){};
+var OCWebPrintTopicEntriesData = function(){};
 // Filestack
 var filestack = {};
 filestack.init = function(){};
--- a/resources/public/lib/print_ascii.js
+++ b/resources/public/lib/print_ascii.js
@@ -7,6 +7,7 @@ OCWebPrintTeamData(): to print the current team data,
 OCWebPrintTeamRoster(): to print the current team roster of users,
 OCWebPrintBoardData(): to print the current board data,
 OCWebPrintEntriesData(): to print all the entries loaded,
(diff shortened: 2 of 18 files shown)


bago2k4:
[Comments UI](https://trello.com/c/aBhDvqzr/826-comments-ui)

belucid:
Reviewing

belucid:
@bago2k4 

I get the same above the topic behavior as Stuart on my laptop:
http://take.ms/s8oIK
In addition, I see those scroll bars (re: the usual Mac setting we see for that problem)

bago2k4:
@belucid pull and try again please. I wasn't able to repeat the problem of the comments box above the topic view but i removed the scroll bars and fixed an issue with the topic view size.
Let me know if this works.

belucid:
@bago2k4 

go to topic A
show comments
then go back to the board
without hiding the comments
then show topic B
comments will be at the top

belucid:
@bago2k4 
if you get enough comments, then when you try to leave a new one, the buttons aren’t visible:
http://take.ms/BXvil

bago2k4:
@belucid good catch, the problem of the comments box showing above the topic view is fixed. Also the other issue. This is good for review again

bago2k4:
@belucid also i'm showing a gradient to fade out the content of the comments, this is happening only if the content exceeds the view height (ie: the comments box scrolls). In your screenshot above it's not visible but it should be. Now it should be fixed, check it out.

PR #257, opened 2017-09-20
Ryan style

bago2k4:
https://trello.com/c/ksnJiI2O


belucid:
Reviewing

belucid:
@bago2k4 down caret on the org picker is supposed to always be there, not just be there on hover, if you have more than 1 org.

belucid:
@bago2k4 moving posts has a flicker.

belucid:
@bago2k4 moved both of these to QA doc.

PR #259, opened 2017-10-04
Change service

belucid:
https://trello.com/c/YpleBO7S
https://trello.com/c/IMTAk8QN

- [x] Merge in mainline and handle conflicts
- [x] Review this web code.

To test this: 

You'll want 3 browsers and 3 users (2 authors, "A" and "B", 1 viewer) and an org w/ a public board, call it "X" and another board, call it "Y".

- [x] For the purpose of this review, you'll want to comment in the line that was commented out here: https://github.com/open-company/open-company-web/pull/259/commits/6397b3c2f9ee30eabca8b117120d24544863ae17
- [x] Log in as user A, and user B in browser window 1 and 2.
- [x] Create some content with A on the board X. Do you NOT see NEW as A on your own newly created content?
- [x] Load board X in the 3rd browser as an anonymous user, do you NOT see any NEW? Good! Anonymous users on public boards don't get any NEWs.
- [x] Login the 3rd browser window now as the viewer, do now you see the new X board content as NEW? Great. Viewers get change services support, but anonymous users do not.
- [x] Create some content with A on board X again. Does the viewer see the content? Is it NEW? Cool.
- [x] Put user B and board Y and user A on board X. Create content with both. Does A see board Y turn bold and the board prefetch happen for board Y. Does B see board X turn bold and the board prefetch happen for board X?
- [x] Navigate user A to board Y. Was the board data already available? Did the bolding of the board name turn off? Did the new post say NEW? Awesome. Now navigate to board X and back to Y. Is the NEW flag gone? Hot damn!
- [x] Navigate user B to board X. Was the board data already available? Did the bolding of the board name turn off? Did the new post say NEW? Awesome. Now navigate to board Y and back to X. Is the NEW flag gone? Sweet!
- [ ] Put user A and user B on the same board (what the heck, put the viewer on the board to). Create some content with A. Do B and the viewer see it immediately'ish? Is it NEW? Humdiggity.
- [x] Do some other thing to posts, create reactions, comments, make updates. Don't see any NEW appearing, correct? Outstanding.
- [x] Think of any other interesting scenarios, test those too! How'd it go?
- [x] Merge
- [x] Learn the banjo

PR #263, opened 2017-10-18
Boards preload

bago2k4:
### Merge into #259 

Card: https://trello.com/c/IMTAk8QN

To test use the card checklist.
Also this fixes a small style issue on `change-service` branch. If you hover on a board that is not the current one but that has new content it will have a light font together with the gray background color at first, then if you keep moving toward the board name it will become bold.

Testing:

- [x] Pre-load other boards
- [x] Reload board on change push
- [x] Open entry modal and exit, no spinner
- [x] Losing NEW after modal open


belucid:
Reviewing.

PR #264, opened 2017-10-19
Settings cancel fix

bago2k4:
This fixes 2 small bugs in invite panel:
- cancel button looks like is floating in the screen
- clicking Cancel looks broken if the invite table is empty

Solution to the second problem is: if there are dirty data clean it, if it's already empty dismiss the view.

### DIFF
--- a/scss/partials/_org_settings.scss
+++ b/scss/partials/_org_settings.scss
@@ -47,15 +47,15 @@ div.org-settings {
 
     div.org-settings-footer {
       margin-top: 64px;
-      height: 90px;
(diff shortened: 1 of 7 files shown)


belucid:
@bago2k4 

- [ ] Now clicking cancel on Team Settings closes the dialog even when there is dirty data. Should do the same as invite and clear on dirty and close on clean.
- [ ] User profile should behave the same for consistency.

belucid:
Code looks good.

bago2k4:
Fixed @belucid

PR #265, opened 2017-10-19
Filestack update

bago2k4:
This rely on filestack update in cljsjs: https://github.com/cljsjs/packages/pull/1371

Once that is merged you can review this.
Right now we have an error if you try to upload an SVG image, that's because filestack API return an error when we try to resize the image to get a thumbnail.
Now we skip the thumbnail creation and we use the just uploaded image as thumbnail too.

To test:
- click New Post
- upload a PNG image
- check that `src` and `data-thumbnail` attributes of the uploaded image are different
- upload an SVG image
- check that you see the uploaded image in the post body
- check that now the 2 attributes are equal and there are no error in the console

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -47,7 +47,7 @@
     [cljsjs/emojione-picker "0.3.6-2"] ; EmojionePicker cljsjs package https://github.com/tommoor/emojione-picker
     [cljsjs/web-animations "2.1.4-0"] ; JavaScript implementation of the Web Animations API https://github.com/web-animations/web-animations-js
     [cljsjs/moment-timezone "0.5.11-0"] ; Timezone support for moment.js https://github.com/moment/moment-timezone/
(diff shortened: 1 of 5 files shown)


belucid:
Looks good.

belucid:
@bago2k4 

- [ ] Now an SVG logo uploads OK since the thumbnail isn't tried, and it displays OK in team settings, but it doesn't display in the header

bago2k4:
@belucid i tried with your SVG image, the problem is that it's bigger than our max allowed width/height and it's not shown in this case (not sure why only SVG, it may have something to do how the browser resize images).
Anyway i fixed this problem, still we have a problem if the user uploads an SVG as his avatar, in that case we try to crop it to a square image but filestack fails (it fails every time you try to crop an SVG image).
Problem is that we need to setup the crop before the user selects the image.

PR #266, opened 2017-10-20
Comment reactions.  Add a single thumbs up emoji to allow the user to "agree" to the comment.

thepug:
Display and allow a user to react to a comment.

https://trello.com/c/lD7AaDcf/836-comment-reactions-agree-thumbs-up

Depends on https://github.com/open-company/open-company-interaction/pull/10

To test
- Make sure you are using the above `comment-reactions` branch for the interaction service.
- Login and add a comment to an entry.
- [x]  Is there a default 👍  emoji next to the comment that is grayed out?
- [x]  Does the gray get removed when hovering over the 👍 ?
- Click the thumbs up.
- [x]  The counter should increase by 1 and the gray color should be removed.
- Click the thumbs up.
- [x]  The counter should go to zero and the 👍  should be gray.
- Click the thumbs up again So that the second test will work.

- Login as a different user.
- Open the entry to see the comment you created with the other user.
- [x] Does it have a count of one with a colored 👍 ?
- Click the 👍 
- [ ] Does the count go to 2?
- Click the 👍 
- [ ] Does the count go to 1?
- [ ] merge


thepug:
I attempted to reuse the reactions component and it is mostly functional.   The flux actions and handlers for those actions assume that the reactions are for the posts (aka entries). Those handlers seem to update data for the entry/post and then try to reload that data.  I need to handle the new case where a reaction is on a comment.

thepug:
Well, it is functional since https://github.com/open-company/open-company-web/pull/266/commits/b0cf9034e90325a0e21072dc20aabaf35de034ea. @bago2k4 when you have some time I'd like you to start to review the code so I can get some feedback on this approach.  

I'll start on the design. It looks like I'll be using 'agree(d)'  instead of 'reacted to' for reactions on comments.
new_post_-_g

bago2k4:
I like the approach, the :reaction-toggle action is growing a bit too much and is doing 2 separate things: one for entries and one for comments, maybe we should use 2 different actions or split that in 2 separate functions, but it's only code smell..

Two things i get from here:
- I (we) should start doing much difference btw functional and presentational (?) components (my bad, i have never used this differentiation before, i need to put more effort in it)
- clean up the code and use only one of entry, activity and post now that stories are gone for good

bago2k4:
Btw @thepug i like a lot that the change to the reactions component is minimal!

thepug:
Design notes from stuart https://opencompanyhq.slack.com/archives/C06SBQ3FD/p1510239078000406

```
I think we plan to use the :thumbsup: instead of the smiley face


[9:51] 
Also, per the latest design tweak from Ryan last night


[9:52] 
I think the icon can be black and white if not clicked. On hover the icon would be in color, and if clicked the bg would get color, too.


[9:52] 
That would match the pattern we’ll be using for reactions on posts
```

thepug:
@bago2k4 , could you clarify the two statements below. Do you want me to address both in this PR?

> Two things i get from here:

> I (we) should start doing much difference btw functional and presentational (?) components (my bad, i have never used this differentiation before, i need to put more effort in it)
clean up the code and use only one of entry, activity and post now that stories are gone for good


I think I understand the second statement and you want that to be done in a future refactor?  Not sure about the first statement.

thepug:
@bago2k4  do we have an example of :arrow_down: anywhere else in the code?  Making an emoji black and white.

> I think the icon can be black and white if not clicked. On hover the icon would be in color, and if clicked the bg would get color, too.

belucid:
technical nit, we actually want it grey scale, not black and white

bago2k4:
@thepug not, sorry, it was just a global takeaway. Sorry i wasn't clear on it.

thepug:
Comments Feedback
- Try reaction a bit smaller
- Fix the half of a thumb icon cutoff
- Pull comment text to the left align with the avatar
- Mirror top and bottom spacing around the agree with the design
- Remove grey-scale when hovering over the button, not the icon
- Don’t have ability to agree to your own comment (should still see other’s agreements though)
- Copy: “You agreed”, “You and +3 agreed”, “3 people agreed”
- Animation shouldn’t be in grayscale
- After agreeing at the top, shouldn’t reset scroll to the bottom
Until you click in the comment box, have it just be 1 line high with “Share your thoughts…”. Once it has focus, it should be the same size as now with the Emoji and Add button
- When setting the modal size, don’t just use how much space the post needs (taking into account the min and max), but check the size that comments need and if it’s bigger than the post size (taking into account the max) then use the comment size.

thepug:
[Comment Reactions (Agree Thumbs up)](https://trello.com/c/lD7AaDcf/836-comment-reactions-agree-thumbs-up)

thepug:
Moved to a new trello card https://trello.com/c/4lTyIrn1/1096-until-you-click-in-the-comment-box-have-it-just-be-1-line-high-with-share-your-thoughts-once-it-has-focus-it-should-be-the-same
> Until you click in the comment box, have it just be 1 line high with “Share your thoughts…”. Once it has focus, it should be the same size as now with the Emoji and Add button

thepug:
> When setting the modal size, don’t just use how much space the post needs (taking into account the min and max), but check the size that comments need and if it’s bigger than the post size (taking into account the max) then use the comment size.

Moved to https://trello.com/c/nXNhN0le/1097-entry-modal-changes-when-comments-are-larger-than-entry-content

belucid:
@thepug not sure if this is meant to be ready or not, but FWIW I'm seeing "1 person agreed" when I agree rather than "You agreed"

thepug:
- [ ] remove agree button if there are no reactions and the logged in user is the owner of the comment.

belucid:
@bago2k4 when this gets merged, merge the Interaction service PR https://github.com/open-company/open-company-interaction/pull/10 It's done, ready to go w/ this one.

bago2k4:
@thepug why there are commits from my `marketing-site-refresh`branch here, or at least it looks like.. i see for example this image /img/ML/home_main_screenshot.png or changes to _about.scss file

thepug:
> @thepug why there are commits from my marketing-site-refreshbranch here, or at least it looks like.. i see for example this image /img/ML/home_main_screenshot.png or changes to _about.scss file

Not sure how that happened @bago2k4   Should be good now.

scss/partials/_comments.scss
```diff
@@ -118,7 +118,7 @@ div.comments {
         font-size: 13px;
         color: rgba(78,90,107,0.8);
         font-weight: 300;
-        margin: 0 0 0 #{$comment-avatar-size + 16}px;
+        margin: 10px 0 10px;
```
bago2k4: Just for later reference, you can use `10px 0;` here, no need to repeat the same at the end.
I'm sure you already know this but 
- `0px`: all 4 values
- `0px 1px`: top/bottom 0px left/right 1px
- `0px 1px 2px`: top 0px, left/right 1px, bottom 2px
- `0px 1px 2px 3px`: top 0px, right 1px, bottom 2px, left 3px

src/oc/web/actions.cljs
```diff
@@ -887,32 +887,72 @@
   (api/get-comments activity-uuid)
   (assoc db :comment-add-finish true))
 
-(defmethod dispatcher/action :reaction-toggle
-  [db [_ activity-uuid reaction-data]]
+(defn- current-board-key
```
bago2k4: Maybe we should move this function to `oc.web.dispatcher`? it already contains all the key functions: `all-posts-key` and `board-data-key` it would make sense to have this there too.
Dispatcher already import `router` so it won't be a problem.
Anyway, your call if you want to move this now or in another branch.

src/oc/web/actions.cljs
```diff
@@ -887,32 +887,72 @@
...
+  (let [activity-id (router/current-activity-id)
         org-slug (router/current-org-slug)
-        board-key (if is-all-posts
-                   (dispatcher/all-posts-key org-slug)
-                   (dispatcher/board-data-key org-slug (router/current-board-slug)))
+        board-slug (router/current-board-slug)
+        comments-key (dispatcher/activity-comments-key org-slug board-slug activity-id)
+        comments-data (get-in db comments-key)
+        comment-idx (utils/index-of comments-data #(= item-uuid (:uuid %)))]
+    (if comment-idx
```
bago2k4: What means if comment-idx is not set? A comment would be helpful here.

src/oc/web/actions.cljs
```diff
@@ -924,10 +964,30 @@
...
+                                   (assoc-in [:reactions reaction-idx] next-reaction-data))]
         (assoc-in db entry-key updated-entry-data)))))
 
+(defn- handle-reaction-to-comment-finish
+  "Update comment data with the reaction data from the API."
+  [db]
+  (let [org-slug (router/current-org-slug)
+        board-slug (router/current-board-slug)
+        activity-uuid (router/current-activity-id)
+        board-key (dispatcher/board-data-key org-slug board-slug)
```
bago2k4: Shouldn't this use `current-board-key` instead? If not why?

src/oc/web/actions.cljs
```diff
@@ -887,32 +887,72 @@
...
+  (let [item-uuid (:uuid item-data)]
+    (api/toggle-reaction item-data reaction-data)
+    (if (= (:content-type item-data) "entry")
+      (handle-reaction-to-entry db item-uuid reaction-data)
+      (handle-reaction-to-comment db item-uuid reaction-data))))
+
+(defn- handle-reaction-to-entry-finish
+  "Update an entry with the reaction data from the API."
+  [db item-data reaction reaction-data]
+  (let [board-key (current-board-key)
```
bago2k4: Shouldn't this use `current-board-key` instead? If not, why?

src/oc/web/components/comments.cljs
```diff
@@ -150,6 +156,15 @@
...
             [:div.comments-internal-scroll
+             {:on-scroll (fn [e]
+                           (let [comments-internal-scroll (js/$ ".comments-internal-scroll")]
+                             ;; when a user scrolls up,
+                             ;; turn off the scroll to bottom after render
+                             (when (and (> (.data comments-internal-scroll "lastScrollTop")
+                                           (.scrollTop comments-internal-scroll))
+                                        @(::scroll-bottom-after-render s))
+                               (reset! (::scroll-bottom-after-render s) false))
+                             (.data comments-internal-scroll "lastScrollTop" (.scrollTop comments-internal-scroll))))}
```
bago2k4: Maybe set `(.scrollTop comments-internal-scroll)` in a var in the let above to avoid it changing value btw the two uses?
Not sure if this is subject to race condition though, your call

bago2k4:
@thepug the last steps of the tests don't work, but i think that's by design since i can't agree to one of my comments, right?

bago2k4:
@thepug other then the comments above the PR is ok

src/oc/web/actions.cljs
thepug: This part doesn't use `current-board-key` because we already have the board-slug. `current-board-key` has a test to see if it the user is reacting from all posts or a single board.

src/oc/web/components/comments.cljs
thepug: It should be the case that the user scrolls and then `data` value is set. This value only changes when the user scrolls again. Since there is only one way to change that value I don't think it should be a race condition.

Am I missing something?

bago2k4: nope, makes sense.

src/oc/web/actions.cljs
bago2k4: right!

bago2k4:
@thepug my tests are failing locally, here is the flow:
- user user A and B that are both at least author of the same org
- publish a post with A and comment on it
- comment on it with B
- agree on the first comment with B: A doesn't see the reaction from B
- agree on the second comment with A: B doesn't see the reaction from A
- refresh both windows and check that the agree are there (+1 on both)

thepug:
@bago2k4, The code to fix the bug you found has been added.

Also the refactor we discussed in chat has been added.  I think other separations (css/actions) will be easy if needed.

bago2k4:
@thepug i'm going to merge this but there is a bug still, explaining it here and adding a trello card too.
You need 3 users for this:
- create a comment with user A and agree with user B
- refresh the post with B and C: B sees You agree and C sees +1
- agree with C
- user B sees: +2 agreed, it lose the `reacted` bool

i think this is because when the notification comes in from the interaction websocket we ignore the current reacted and we replace the record.
It should be quick to fix though

PR #268, opened 2017-10-23
Preload CSS images

bago2k4:
Add a preload images SCSS mixin that can be used to preload whatever image via CSS.
For now it preloads only the white plus icon used in the add board button.

To test:
- open an incognito tab or clean up all your browser cached data (better the first to make sure)
- signup
- create an org
- when you land on the welcome board check in the network tab if the image called `plus_button_icon_white.svg` is already loaded, was it? good
- follow the onboard and stop at the second bat signal tooltip, do you see the white plus appearing with the tooltip? good

### DIFF
--- a/scss/mixins/_reusable.scss
+++ b/scss/mixins/_reusable.scss
@@ -243,3 +243,25 @@
     background-image: cdnUrl($image + "@2x" + $suffix);
   }
 }
+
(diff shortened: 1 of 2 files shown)


belucid:
Reviewing.

PR #269, opened 2017-10-24
Change service fixes

bago2k4:
This PR is needed because `change-service` PR #259 has a couple of issues, the first is about this item in the test steps:
> Put user A and user B on the same board (what the heck, put the viewer on the board to). Create some content with A. Do B and the viewer see it immediately'ish? Is it NEW? Humdiggity.

And that's fixed by the first commit here, to test open the same board X with 3 users: 2 authors and a viewer and with the first of them add a post to X. Does the board autoupdate for user 2 and 3? Good.

The second issue, fixed with the second commit, it's that on firefox email addresses with uppercase letters will add the red error border around the email field in: signup with email, invite by email and login with email.
To test try to put this testEMAIL@ADDress.com one or all the 3 the specified places.


belucid:
Reviewing.

PR #271, opened 2017-10-26
Jelly

bago2k4:
New jelly design and general improvements

Card: https://trello.com/c/tXlzcK0G

PR #272, opened 2017-10-26
Native emojis

bago2k4:
## NB: review after #271 

This is needed to resolve the problems emoji in headline are causing in Safari and FireFox.

Discussion happened in Slack and Talky but we finally decided that it was better to go with the native emojis for a number of benefits and one major problem: the picker still uses the image emojis.
We will fix this in a second moment most probably using this https://github.com/missive/emoji-mart or something close to it.

Card: https://trello.com/c/uVaZbfmZ


belucid:
Reviewing.

@bago2k4 Just going to handle this as a PR into Jelly instead.

PR #275, opened 2017-11-06
Code cleanup

bago2k4:
Clean code thanks to yagni, eastwood, kibit and bikeshed.
Add the `check-sources` task that runs all the above plugins to travis configutaion.


belucid:
Reviewing.

PR #276, opened 2017-11-08
All posts improvements

bago2k4:
Slack discussion, Stuart said:
> @here if you open up All Posts you’ll see the Boards icon in the board label. That grid icon used to make more sense because we also showed it in navigation; but not any more.

> I think it can be confusing. I’m always a little surprised by it when I go into AP because I don’t see it anywhere else. 

> Why not just add the board name to the time stamp, e.g. “1 day ago in Design” or “October 29 in Product Feedback”? Then we’d only have the Topic pillbox if there is one which is more consistent.

I added the link to the board in bold.

Tests:
- [ ] Review the code
- [ ] Open All Posts, do you NOT see the icon besides the link in left nav? Good
- [ ] Do you NOT see the board name in the boards top right? Good
- [ ] Do you see the board name in bold besides the post creation/update date? Good
- [ ] Is it clickable and sends you to the last viewed sort of that board? Good
- [ ] Merge
- [ ] Do one of Sean's things when you close a PR

### DIFF
--- a/scss/mixins/_activity.scss
+++ b/scss/mixins/_activity.scss
@@ -29,6 +29,17 @@ $h2_top_margin: 20;
     font-size: 12px;
     white-space: nowrap;
 
+    time a.board-name {
+      text-decoration: none;
+      @include avenir_H();
+      color: $carrot_light_gray_3;
+      font-size: 12px;
+
+      &:hover {
+        text-decoration: underline;
+      }
+    }
+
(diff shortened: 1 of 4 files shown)


bago2k4:
We decided on Slack to not go through with these changes and to simply remove the board icon form the AP pillboxes. Opening a new PR for that.

PR #278, opened 2017-11-08
Remove board icons form All Posts pillboxes.

bago2k4:
From Stuart in Slack:
> so the only real change is putting the AP icon back, and taking the boards icon off

...from the AP pillboxes

To test:
- [x] Check that the icon is gone from AP posts pillboxes


belucid:
@bago2k4 I'm thinking we want the board pillbox to hover in blue rather than purple. Purple is the color we use for topics. We don't REALLY have a color that means boards, but we use blue in the left nav. for boards, so that seems right.

PR #282, opened 2017-11-08
Post drafts

bago2k4:
Card: https://trello.com/c/ubX7kXJu

Mostly test these:
- Create draft
- Update existing draft
- Publish draft (check the date that appear on the board card once it's published, is it the publish date? Good
- Update the published post
- Publish a post without saving to drafts
- Update the directly published post
- Move a draft btw boards
- Change a draft topic
- Change a published post topic
- Move a draft from board A to board B with a topic that didn't not exist on board B. Is it maintained? Good
- Move a published draft btw boards
- Change topic of a published post
- Start post while you are filtering on a topic, does the new draft/post inherit the topic? Good

NB: this is missing a couple of items from the QA doc


bago2k4:
[Drafts for Posts](https://trello.com/c/ubX7kXJu/1090-drafts-for-posts)

belucid:
Reviewing.

belucid:
@bago2k4 

`dashboard_layout.cljs?rel=1510667557121:31 Uncaught TypeError: Cannot read property 'scrollTop' of null`

With 5 or 6 drafts.

belucid:
@bago2k4 update dialog copy when creating a new post (not when editing a draft post) to just be "Posting in " rather than "Draft for ". Leave "Draft for " only for when editing a draft.

belucid:
@bago2k4 when I create a new post, the dropdown list of topics is populated and correct for the board. When I edit an existing post, the dropdown list of topics is empty.

belucid:
@bago2k4 have to scroll the dropdown list here (has defaults + 1 custom). Seems the dropdown should be sized so no scrolling is needed here, y?

bago2k4:
@belucid firsts 3 issues addressed.
About the last one: i had to set a maximum height since sometime the topics dropdown was exceeding the window height. I set it static and not depending on the window height because the dropdown is not always in the same spot so it's hard to know how much space it has below.
It depends from the height of the modal that depends from the content of the post.
I'll remove this limit now so you can test. Add 4 custom topics and check on an empty post what happens.

belucid:
All good now.

PR #287, opened 2017-11-16
Comment box resize.

thepug:
Until you click in the comment box, have it just be 1 line high with “Share your thoughts…”. Once it has focus, it should be the same size as now with the Emoji and Add button

To test:

- open a post with comments
- click the "Share your thoughts..."
- [x] Does the input resize and show the emoji and add buttons?
- [x] Does adding a comment resize it back to the smaller version?
- [x] Does closing the post also resize it after reopening the closed post?

merge

### DIFF
--- a/scss/partials/_comments.scss
+++ b/scss/partials/_comments.scss
@@ -34,7 +34,7 @@ div.comments {
     top: 0;
 
     &.add-comment-focus {
-      padding-bottom: 148px;
+      height: calc(100% - 123px);
     }
 
     div.comments-internal-scroll {
@@ -152,7 +152,7 @@ div.comments {
 
       left: 0;
       right: 0;
-      bottom: 12px;
+      bottom: 0px;
       width: 100%;
       margin-top: 16px;
 
--- a/scss/partials/_comments.scss
+++ b/scss/partials/_comments.scss
@@ -25,13 +25,17 @@ div.comments {
 
   div.comments-internal {
     min-height: 127px;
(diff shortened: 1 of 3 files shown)


thepug:
[Until you click in the comment box, have it just be 1 line high with “Share your thoughts…”. Once it has focus, it should be the same size as now with the Emoji and Add button](https://trello.com/c/4lTyIrn1/1096-until-you-click-in-the-comment-box-have-it-just-be-1-line-high-with-share-your-thoughts-once-it-has-focus-it-should-be-the-same)

bago2k4:
Reviewing

bago2k4:
@thepug why are you using the app-state to keep `:add-comment-focus`? it looks like it's used only in comments component and it's reset to false every time the component is mounted.
Seems like the component internal state should be enough.

thepug:
> @thepug why are you using the app-state to keep :add-comment-focus? it looks like it's used only in comments component and it's reset to false every time the component is mounted.
Seems like the component internal state should be enough.

Good catch, this was left over for when I was going to have the box resize back on an outside click.
I'll change this to local.

bago2k4:
Cool, other than that i'm ok with merging this, let me know when you push these changes..

thepug:
Ah, so it turns out since I use the `:add-comment-focus` in the app state across components. It can't be local because of this.  `add-comment` and `comments` are the component both defined in `comments.cljs`.

PR #291, opened 2017-11-27
Fix bold

bago2k4:
Fix CMD+B on medium editor. There are 3 players here: chrome default CMD+B that wraps highlighted text into a B tags, medium editor default extension that use it's own bold (that we disabled for the toolbar), our own bolding that didn't have a CMD+B handling before.
Even disabling the medium editor handling the chrome (or browser) bold handling was taking place and it was applying the default bolding, not our custom.
So i had to add the CMD+B handling for our custom bolding so we apply it how we want and block the event before it bubbles up to medium editor or the browser itself.

To test check that the problems in these looms are solved:
https://www.useloom.com/share/b6a370abbbb74d13b78aaed6a1dc0b39
https://www.useloom.com/share/95311d6acca643f0a36b80165373958b

Test the bold formatting from the toolbar to make sure it works and mix the 2 bolding methods. Also test that CMD+i and CMD+U are not broken. Since we don't provide underline in the toolbar we may want to disable underline via keyboard too

This also disable the image dragging default medium editor extension since it doesn't uses FileStack upload but it simply put images inline.

### DIFF
--- a/resources/public/lib/MediumEditorExtensions/MediumEditorCustomBold/CustomBold.js
+++ b/resources/public/lib/MediumEditorExtensions/MediumEditorCustomBold/CustomBold.js
@@ -1,29 +1,47 @@
+rangy.init();
+
 var CustomBold = MediumEditor.extensions.button.extend({
   name: 'custombold',
 
-  tagNames: ['b'],
+  tagNames: ['b', 'strong'],
   contentDefault: '<b>B</b>',
-  aria: 'Bold',
-  action: 'bold',
+  contentFA: '<i class="fa fa-bold"></i>',
+  aria: 'bold',
+  action: 'custombold',
+  style: {
+    prop: 'font-family',
+    value: 'AvenirLTStd-Heavy, Muli, Arial, sans-serif'
+  },
+  key: "B",
 
   init: function () {
-    rangy.init();
-
     MediumEditor.extensions.button.prototype.init.call(this);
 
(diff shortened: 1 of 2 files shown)


belucid:
Reviewing.

belucid:
I guess what's problematic from a UX standpoint @bago2k4 is CMD-B bolds, but it doesn't unbold (a selection), and it doesn't even turn off bolding. That feels broken. Start to type something, type CMD-B to turn on bolding, type the bolded words, then type CMD-B to turn off bolding and continue on, you expect that much to work.

CMD-i works as you'd expect for italics. Both turning off italics for future typing and the selection.

Also, we should disable CMD-U.

belucid:
Also, FWIW, CMD-B in FireFox is triggering bookmark window, rather than bold. Googling about, it seems we have some control of this behavior with the HTML. Worth looking into if it's not long to fix.

bago2k4:
Closing for the problems above, it looks like it's not worth spending all this time to fix this. We will try using Avenir Book with 700 weight instead of Avenir Heavy with 400/normal weight.
The point is the weight, both medium editor and the browser (with queryCommandState) use the weight to check if bold is already applied.

bago2k4:
[Ongoing UX and Bugs misc](https://trello.com/c/YRFu6ekU/1021-ongoing-ux-and-bugs-misc)

PR #292, opened 2017-11-27
Marketing site refresh

bago2k4:
Card: https://trello.com/c/UbcA7r3F

This is already deployed to staging with some other work in `merge-branch`. Beta has an older version of this so you need to deploy master there once this is merged.

Carrot.io (marketing site) has another version of this work contained in `marketing-site-refresh+try-it-form` since we don't need login/signup/get started buttons there but we need the collect email forms.
Once this is merged you need to replace `marketing-site` branch (that's the one we usually deploy to carrot.io) with the branch mentioned above (`...+try-it-form`) and redeploy.


src/oc/web/components/ui/site_header.cljs
```diff
@@ -3,6 +3,7 @@
    and every change here should be reflected there and vice versa."
   (:require [rum.core :as rum]
             [dommy.core :as dommy :refer-macros (sel1)]
+            [org.martinklepsch.derivatives :as drv]
```
thepug: Not sure why I am getting this locally?
On staging I get the same but it is `app.main.css`

screenshot

PR #293, opened 2017-11-28
Comment delete

thepug:
Trello: https://trello.com/c/ClMlhLsU/1105-comments-ui-delete

This change adds a drop down menu on comments owned by the current user. The menu has a 'Delete' option that will prompt the user if they are sure they want to delete when clicked. Once clicked the comment is removed.

To test:

- Open up an activity with comments owned by your current test user.
- [x] Hover over the comment and a trash can will appear on the top right of the comment.
- [x] Clicking the delete can will bring up a modal asking if you are sure you want to delete.
- [x] Clicking yes will remove the comment.

- Login as a different user and leave the first user logged in.
- Add a comment to the same activity the first user is viewing.
- Follow the same delete steps as above.
- [x] Is the comment removed from both user's views?
- merge

### DIFF
--- a/src/oc/web/actions.cljs
+++ b/src/oc/web/actions.cljs
@@ -900,7 +900,7 @@
 
 (defmethod dispatcher/action :comment-delete
   [db [_ activity-uuid comment-data]]
-  (api/delete-comment comment-data)
+  (api/delete-comment activity-uuid comment-data)
   (let [org-slug (router/current-org-slug)
         board-slug (router/current-board-slug)
         item-uuid (:uuid comment-data)
@@ -910,8 +910,16 @@
     (assoc-in db comments-key new-comments-data)))
 
 (defmethod dispatcher/action :comment-delete/finish
-  [db [_]]
-  db)
+  [db [_ {:keys [success activity-uuid]}]]
+  (if success
+    db
+    (let [org-slug (router/current-org-slug)
+          board-slug (router/current-board-slug)
+          board-key (dispatcher/board-data-key org-slug board-slug)
+          board-data (get-in db board-key)
+          activity-data (get-in board-data [:fixed-items activity-uuid])]
+      (api/get-comments activity-data)
+      db)))
 
 (defmethod dispatcher/action :ws-interaction/comment-delete
   [db [_ comment-data]]
--- a/src/oc/web/actions.cljs
+++ b/src/oc/web/actions.cljs
@@ -894,9 +894,36 @@
     (assoc-in db comments-key new-comments-data)))
 
 (defmethod dispatcher/action :comment-add/finish
-  [db [_ {:keys [activity-uuid]}]]
-  (api/get-comments activity-uuid)
+  [db [_ {:keys [activity-data]}]]
+  (api/get-comments activity-data)
   (assoc db :comment-add-finish true))
+
+(defmethod dispatcher/action :comment-delete
+  [db [_ activity-uuid comment-data]]
+  (api/delete-comment comment-data)
+  (let [org-slug (router/current-org-slug)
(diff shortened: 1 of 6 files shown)


thepug:
[Comments UI - Delete](https://trello.com/c/ClMlhLsU/1105-comments-ui-delete)

thepug:
- [ ] Replace ellipsis with trash can. Click on trash can starts the delete process.

thepug:
@bago2k4 ready for review again

bago2k4:
reviewing

src/oc/web/api.cljs
```diff
@@ -692,7 +692,18 @@
           (dispatcher/dispatch! [:comment-add/finish {:success success
                                                       :error (when-not success body)
                                                       :body (when (seq body) (json->cljs body))
-                                                      :activity-uuid (:uuid activity-data)}]))))))
+                                                      :activity-data activity-data}]))))))
+
+(defn delete-comment
+  [comment-data]
+  (when (and comment-data)
```
bago2k4: @thepug i dont' think you need `and` here

src/oc/web/actions.cljs
```diff
@@ -894,10 +894,37 @@
...
+        board-slug (router/current-board-slug)
+        item-uuid (:uuid comment-data)
+        comments-key (dispatcher/activity-comments-key org-slug board-slug activity-uuid)
+        comments-data (get-in db comments-key)
+        new-comments-data (remove #(= item-uuid (:uuid %)) comments-data)]
+    (assoc-in db comments-key new-comments-data)))
+
+(defmethod dispatcher/action :comment-delete/finish
+  [db [_]]
+  db)
```
bago2k4: @thepug I see you send the `:success` value with `:comment-delete/finish` from oc.web.api. Maybe we can reload the comments if comment delete failed?
Since you have already removed the comment from the app-state object.
Every time a 5xx or a 0 is received from an api call a red banner is shown at the top of the page. With this we only need to make sure that the app-state is still in sync with the server.
If you don't like reloading comments you can wait to remove the comment from the app-state to here when you receive the response. Your call.

src/oc/web/components/comments.cljs
```diff
@@ -25,7 +47,18 @@
           [:div.comment-author
             (:name author)]
           [:div.comment-timestamp
-            (utils/time-since (:created-at c))]]]
+            (utils/time-since (:created-at c))]]
+          (when (seq (:links c))
```
bago2k4: @thepug  i think you should check if `(utils/link-for (:links c) "delete")` is present instead of checking if there are links. This may be a problem if more links are added.

src/oc/web/components/comments.cljs
```diff
@@ -14,8 +18,26 @@
...
+                    :action "delete-comment"
+                    :message (str "Delete this comment?")
+                    :link-button-title "No"
+                    :link-button-cb #(dis/dispatch! [:alert-modal-hide])
+                    :solid-button-title "Yes"
+                    :solid-button-cb #(let [org-slug (router/current-org-slug)
+                                            board-slug (router/current-board-slug)
+                                            board-url (utils/get-board-url org-slug board-slug)
+                                            activity-uuid (router/current-activity-id)]
+                                       (router/nav! (oc-urls/entry org-slug board-slug activity-uuid))
```
bago2k4: @thepug why do you navigate to the entry again? you are already here, maybe to force comments reload?

thepug:
@bago2k4 issues addressed in the last commit. bf2e7ed

bago2k4:
@thepug as discussed in Slack it needs only the trash icon update, we need to use the material design icon for delete. Other than that this can be merged

PR #294, opened 2017-11-28
New reactions

bago2k4:
Card: https://trello.com/c/jRmI3Ta8 (second part, emoji-mart picker was already done)

Some general testing:
- setup 3 authors (or admin) under the same org
- create a board and a post
- pick 1️⃣  with user A, check the other 2 receive it
- check that B and C see the reaction
- pick 😂 emoji with B
- check that A and C see the reaction
- react with C to one of the previous and check that both A and B see it
- now un-react with A and B to an emoji and bring it back
- make sure that when all the reaction are removed the emoji is still present but no number is shown
- make sure that you can always add and remove reactions when that reaction changes live
- refresh the page and make sure you still see the same as before (the empty reactions may be gone)
- now agree to a comment and check that you see it with the other 2 users and no error is shown
- refresh and check that the comment agree is still there
- disagree to the comment

Make all the tests you can think of using 3 users.


belucid:
Reviewing.

bago2k4:
@belucid Ready again

belucid:
@bago2k4 Issue when you react w/ a favorite, it animates and posts, but once the animation is done, doesn't turn it to active.

Easiest way to test:
- make sure you are running latest `new-reactions` storage branch
- create a brand new post
- notice you have 3 reactions (presumably, if you or your org has reacted w/ at least 3 different reactions before)
- react with one of these, rather than with the picker
- notice you get the animation, but react stays inactive
- but notice the PUT worked and responded same response as the POST from the picker, and if you return to the board or refresh, your reaction is there

thepug:
[Emoji-mart and reactions picker](https://trello.com/c/jRmI3Ta8/1098-emoji-mart-and-reactions-picker)

belucid:
Reviewing again.

belucid:
QA tests all check out.

src/oc/web/actions.cljs
```diff
@@ -1693,6 +1697,37 @@
       (assoc-in db dispatcher/whats-new-key fixed-whats-new-data))
     db))
 
+(defmethod dispatcher/action :react-from-picker
```
belucid: action should be in README

src/oc/web/actions.cljs
```diff
@@ -1693,6 +1697,37 @@
...
+  [db [_ activity-data emoji]]
+  (when (and emoji
+             (utils/link-for (:links activity-data) "react"))
+    (api/react-from-picker activity-data emoji))
+  db)
+
+(defmethod dispatcher/action :react-from-picker/finish
```
belucid: action should be in README

bago2k4:
@belucid added the missing actions, should i merge? would you?

belucid:
@bago2k4 This is ready to go but we'll wait to merge until the back end PRs are ready to go out.

PR #297, opened 2017-11-30
Fix small UI bug where you can interact w/ channel picker when disabled.

thepug:
Trello: https://trello.com/c/wKY660ui/1110-fix-small-ui-bug-where-you-can-interact-w-channel-picker-even-when-its-toggled-off

Also Fix small UI bug, no tooltip on the informational i in the dialog.

To test:

- Create a new board with a team that has slack enabled.
- In the new board UI you will see a toggle for Slack and make sure it is off.
- [x] You should not be able to click the down arrow for choosing a slack channel.

- [x] Mouse over the i icon to see the tooltip

### DIFF
--- a/src/oc/web/components/board_edit.cljs
+++ b/src/oc/web/components/board_edit.cljs
@@ -167,7 +167,10 @@
               [:div.board-edit-slack-channels-label.group
                 [:div.title
                   "Connect comments to Slack"
-                  [:span.more-info]]
+                  [:span.more-info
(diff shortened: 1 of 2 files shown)


thepug:
[Fix small UI bug where you can interact w/ channel picker even when it's toggled off.](https://trello.com/c/wKY660ui/1110-fix-small-ui-bug-where-you-can-interact-w-channel-picker-even-when-its-toggled-off)

PR #298, opened 2017-11-30
Fix bold

bago2k4:
Fix bold issue by using Avenir Book with 700 weight in all body. Also add some minor fixes from the QA doc: https://docs.google.com/document/d/1X3j869Y8n94zfqaGVhrWf8VOlr3qEa_4a1fxtIjH6fs/edit#


bago2k4:
[Ongoing UX and Bugs misc](https://trello.com/c/YRFu6ekU/1021-ongoing-ux-and-bugs-misc)

bago2k4:
Bug: on FireFox the emoji picker is always disabled when focused on the body. The problem is that FF doesn't trigger FOCUSIN and FOCUSOUT events at document level when a field in the page gets focused in or out so the emoji button is always disabled.

Proposed fix: on FF only listens for all window CLICK or KEYPRESS and check that the current `activeElement` has the `.emojiable` class, we do the same right no document FOCUSIN and FOCUSOUT.

belucid:
Reviewing again.

belucid:
plain and italic don't play well together @bago2k4 

I can eventually get them to work with much futzing and see bold italic: http://take.ms/bZLFB

But it's not easy if you just highlight something and click bold and italic it doesn't work, or if you just type command B and command I one after the other and then type it doesn't work. I'm not entirely certain actually what does make it work, but you can eventually get bold italic by futzing with it.

Not a showstopper so moved it to the list rather than hold up this PR. Rest seems good.

PR #300, opened 2017-12-04
if slack is enabled, make slack the default share option.

thepug:
Trello: https://trello.com/c/Kz3F6DME/1108-tighter-slack-board-integration

The final change in the above trello card.  This change makes slack the default when manually sharing an entry.

To test:

- Login with a team that has slack enabled.
- Configure a board to share to slack.
- Click share on an entry.
- [x] Is slack the default instead of url?

- Login without slack enabled.
- Click share on an entry.
- [x] Is slack greyed out and url the default?


thepug:
[Tighter Slack -> Board integration](https://trello.com/c/Kz3F6DME/1108-tighter-slack-board-integration)

PR #302, opened 2017-12-07
Preload EmojiMart picker

bago2k4:
Card: https://trello.com/c/tJ1K6Ywe

Preload the EmojiMart picker when the emoji-picker component is added so there is no delay when the user clicks on the picker button.

To test:
- click New post button
- [x] click the emoji button as fast as you can, do you see the picker? does it show immediately (less then 1 sec delay)? Good
- [x] repeat the test above for a post edit clicking the picker button at the bottom of the  post
- [x] repeat clicking the picker button in the comments

### DIFF
--- a/scss/partials/_emoji_picker.scss
+++ b/scss/partials/_emoji_picker.scss
@@ -37,6 +37,12 @@ div.emoji-picker {
     position: absolute;
     display: none;
 
+    &:not(.visible).preloading {
+      display: block;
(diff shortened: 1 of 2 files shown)


bago2k4:
[Preload emoji picker to avoid delay on first click](https://trello.com/c/tJ1K6Ywe/1119-preload-emoji-picker-to-avoid-delay-on-first-click)

PR #309, opened 2017-12-14
Fix Login Overlay

bago2k4:
This is meant to fix a UI/UX annoying problem.

When you try to login, after typing the first letter in the email or password field the cursor is moved back to the initial value. This is because the app-state field that keep that value is nil so React re-render the input field after you start typing and the cursor is moved.

This should fix that issue.

To test simply open the login view in incognito e start typing the email, can you type the email w/o having to move the cursor back to the end of the field? Good


thepug:
[When logging in the input fields reset while typing.](https://trello.com/c/8q2DFecz/1117-when-logging-in-the-input-fields-reset-while-typing)

PR #310, opened 2017-12-15
Private board invite

bago2k4:
Card: https://trello.com/c/RnKiS5Fl

To test:
- using a user that is logged in with slack
- create a private board
- check that you can add and remove users and change permission of the private board
- check that you can't add slack user that didn't join the carrot team yet
- now change the board permissions to public
- refresh the window
- open board settings again
- do you NOT see a UI jump in the view? Good
- change board permissions back to private
- do you still see the list of users you setup before? Good
- make any other test you can think of with a private board


belucid:
@bago2k4 let's go ahead and use this same PR/branch for the batch add/remove.

belucid:
@bago2k4 this checks out very well against the storage PR for batching.

Only issue is "Leave Board" no longer does anything except close the dialog. It should probably still use the old 1 off real-time remove member link.

belucid:
@bago2k4 Ready to go after that last fix is in and reviewed.

PR #311, opened 2017-12-17
Misc punchlist

bago2k4:
When i was creating a post today on beta i found out some small issues that i fixed here.
These are the issues:
1. create new post modal is not re-centering itself when the browser window resize
2. h2 titles as first line of body are bigger then they should
3. h2 titles in FoC (of drafts too) are using a bigger font then they should (they are h2 but they should look like normal body)
4. if you have a local save of an edited post and start creating a post on another board it lose the board name and slug you wanted it on
5. if you save to draft a post that was already a draft you don't see your last edits until you refresh

Most of them are very simple and don't require additional test steps. But number 4 needs testing:
- hit new post on board A
- add a title and body (simple that you can remember)
- dismiss the modal and click cancel without saving (so they are saved locally)
- switch bot board B
- hit new post here
- [ ] do you see board B up top? Good
- [ ] can you save to draft? Good
- [ ] can you post (please from the initial post creation)? Good

Tests for number 5:
- hit new post
- add only a title (no body for the moment)
- save to drafts
- open the draft to edit it
- add a body
- save to drafts again
- [ ] Do you see the added body in the FoC of the Drafts board? Good
- [ ] Can you post it? Good
- [ ] Do you see it on the board it should be on? Good
- [ ] Do you not see it anymore on the drafts board? Good

### DIFF
--- a/scss/mixins/_activity.scss
+++ b/scss/mixins/_activity.scss
@@ -157,6 +157,9 @@ $h2_top_margin: 20;
 
   h2:not(:first-child) {
     margin-top: #{$bottom_margin + $h2_top_margin}px;
+  }
+
+  h2 {
     @include avenir_H();
     font-size: #{$font-size + 3}px;
 
--- a/scss/partials/_activity_card.scss
+++ b/scss/partials/_activity_card.scss
@@ -347,6 +347,12 @@ div.activity-card {
         display: none;
       }
 
+      h2 {
+        @include avenir_R();
+        font-size: 16px;
+        font-weight: normal;
+      }
+
       b, strong {
         font-weight: normal;
         @include avenir_R();
--- a/scss/partials/_drafts_layout.scss
+++ b/scss/partials/_drafts_layout.scss
@@ -81,6 +81,12 @@ div.drafts-layout {
               @include avenir_M();
             }
 
+            h2 {
+              @include avenir_R();
+              font-size: 16px;
+              font-weight: normal;
+            }
+
             a.read-more {
               @include avenir_R();
               font-size: 16px;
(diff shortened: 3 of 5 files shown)


belucid:
Code looks good.

belucid:
@bago2k4 isn't re-centering vertically on FF http://take.ms/y5AJr

Interestingly, it does if I trigger the "Cancel without saving your changes?" dialog with the X.

bago2k4:
@belucid looks like FF has a more aggressive cache of Chrome, i added some logs and refreshed a few times to check on the re-center issue and now it works. Make sure you have cache disabled in Network tab (it's enabled by default, in Chrome it's not i think).

Also i added a fix for the first draft ever issue. To test on it:
- signup with a new user
- finish NUX
- create a new post to add to the only board  you have
- save as draft
- [ ] make sure you are redirected to the Drafts board
- [ ] make sure the board name is there
- [ ] make sure the board settings button is NOT visible
- [ ] make sure the 🔒 Private label is visible
- [ ] make sure you can edit the draft from here
- [ ] make sure you can post the draft from here

belucid:
@bago2k4 we don't want to do a local save if they agree to lose their changes in the modal.

belucid:
@bago2k4 when I follow the steps above to verify the drafts board for new user fix, I see a brief flash of the blue placebo screen as it navigates from All Hands to Drafts when I click "Save as Draft"

bago2k4:
@belucid local autosave every 5 seconds is in place. Now it asks before leaving the page only if you are in this 5 seconds range where edits aren't saved.
Ready for review again

PR #312, opened 2017-12-21
Awesomeness grab bag

bago2k4:
Card: https://trello.com/c/c67IqQSi

Details about these changes are in [the GDoc](https://docs.google.com/document/d/1X3j869Y8n94zfqaGVhrWf8VOlr3qEa_4a1fxtIjH6fs/edit)


bago2k4:
[Awesomeness grab bag v2.0](https://trello.com/c/c67IqQSi/1131-awesomeness-grab-bag-v20)

resources/public/js/static-js.js
```diff
@@ -20,24 +20,10 @@ function OCStaticMailchimpApiSubmit(e, form, success, fail){
 }
 
 function OCStaticGetCookie(name) {
```
thepug: @bago2k4 can you leave some comments on how this function is used? 

My concern here is how much the code has changed and the fact that decodeURI is no longer used. Anyway something to clear up the functions use will be good.

bago2k4: Good point about decodeURI not used, i'll add it.
For the missing comment i thought the name was pretty self explanatory: it's used to get a single cookie value given its name.
I'll add the comment too though

PR #316, opened 2018-01-02
Search

thepug:
Adds a search box in the top left and uses the search service to find results of posts.

To test:

- make sure search service is running locally
- make sure some data is indexed by editing or creating a post.
- input more than one character in the search box
- [x] does the empty result spy glass image display?
- [x] does clicking outside of the search box remove the spy glass?
- [x] does clicking inside the search box redisplay the spy glass?
- input a search term that shows results
- [x] does the expected result get displayed?
- click on result
- [x] does the post open?

- [x] merge
- [x] before deploying to beta make sure we have a successfully deployed search service to beta
- [ ] deploy to beta


thepug:
[Full text search](https://trello.com/c/x66ILov4/802-full-text-search)

bago2k4:
@thepug it does what it says, one small UI fix maybe, not sure this is the right place though since there is lot of UI improvements to do yet.
If i have search results and click out, i see the results dropdown going away, when i click back into the search field the empty results is shown briefly and then the results come back.
Is this because we have to repeat the search? if so why don't we:
- show a spinner while searching (more UI related maybe)
- keep the results until the search field changes (this is maybe more in the scope of this PR)

Let me know WYT

Other than this it looks good.

thepug:
> keep the results until the search field changes (this is maybe more in the scope of this PR)

@bago2k4 , I'll take a quick go at keeping the results.

bago2k4:
@thepug can you solve the mainline conflict please?

thepug:
> @thepug can you solve the mainline conflict please?

done

PR #317, opened 2018-01-02
Comments autolink

bago2k4:
Card: https://trello.com/c/LyZOf6zw

From a discussion in Slack with Stuart we decided to adopt MediumEditor even for our new and edit comment fields. This means that we can use its autolink feature that automatically turn links into anchor and the paste one that clean all the rich HTML before pasting content in the field.
All the rest of the features are disabled.

To test:
- focus the add comment link
- press Escape: do you NOT see the modal dismiss and go back to the board view? Good
- focus on the add comment and blur it while the field is empty, does it collapse back? Good
- type something with a url in it and check that it's turned into a link, it works? Good
- select some text, do you NOT see the MediumEditor toolbar appearing? Good
- while typing use CMD+B or CMD+I or CMD+U, do you NOT see the bold, italic or underline formatting appear when you type more? Good
- add the comment, does the link is still clickable in the new added comment? Good
- now from the horizontal ellipsis of the newly added comment click edit
- press Escape while the comment is in edit mode, do you NOT see the modal dismiss and go back to board? Good
- edit again, now click outside the comment, is it NOT the comment still editable? Good
- repeat the above tests of the add comment for the comment edit

NB: archive this card https://trello.com/c/BZ0fA4Rm/1113-auto-link-urls-in-comments when done since it's a duplicate of this one but with less details

### DIFF
--- a/src/oc/web/components/comments.cljs
+++ b/src/oc/web/components/comments.cljs
@@ -65,8 +65,9 @@
 (defn cancel-edit
   [e s c]
   (.stopPropagation e)
-  (let [comment-field (rum/ref-node s "comment-body")]
-    (set! (.-innerHTML comment-field) (utils/emojify @(::initial-body s) true))
+  (let [comment-field (rum/ref-node s "comment-body")
+        comment-data (first (:rum/args s))]
+    (set! (.-innerHTML comment-field) (utils/emojify (:body comment-data) true))
     (stop-editing s)))
 
 (defn add-comment-content [add-comment-div]
@@ -144,7 +145,6 @@
                          (rum/local false ::medium-editor)
                          (rum/local nil ::window-click)
                          (rum/local false ::show-more-dropdown)
-                         (rum/local false ::initial-body)
                          (rum/local nil ::esc-key-listener)
                          {:did-mount (fn [s]
                             (reset! (::window-click s)
@@ -159,8 +159,6 @@
                                 (when (and @(::show-more-dropdown s)
                                            (not (utils/event-inside? e (rum/ref-node s "comment-edit-delete"))))
                                   (reset! (::show-more-dropdown s) false)))))
-                            (let [comment-data (first (:rum/args s))]
-                              (reset! (::initial-body s) (:body comment-data)))
                             s)
                           :will-unmount (fn [s]
                             (events/unlistenByKey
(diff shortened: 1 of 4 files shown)


bago2k4:
[Comments autolink and paste cleanup.](https://trello.com/c/LyZOf6zw/1139-comments-autolink-and-paste-cleanup)

src/oc/web/components/comments.cljs
```diff
@@ -19,39 +19,112 @@
...
+      (reset! (::esc-key-listener s) nil))
+    (reset! (::medium-editor s) nil)
+    (reset! (::editing? s) false)
+    (dis/dispatch! [:input [:comment-edit] nil])))
+
+(defn cancel-edit
+  [e s c]
+  (.stopPropagation e)
+  (let [comment-field (rum/ref-node s "comment-body")]
+    (set! (.-innerHTML comment-field) (utils/emojify @(::initial-body s) true))
```
thepug: I am seeing a bug and I don't think you want `::initial-body` here. To recreate the issue edit a comment and save with enter.  After that open the comment and hit escape. You will see the original data.

What you want here is to use the current comment data. This is where you will want to read data from the "Store" to get the current state. Right now that means read from the app state.

bago2k4: @thepug addressed the issue, good catch!

thepug:
One more issue.

Open a post.  Click the "Add comment" box.  Click escape.  Click the "Add comment" box again and you are unable to post a comment.

thepug:
Now I don't get the 'Add' button after typing. 

http://recordit.co/m4eMtmLGyq
m4emtmlgyq

bago2k4:
@thepug good catch again, my bad i introduced a regression while fixing the last issue you found. Now it should be good

thepug:
> @thepug good catch again, my bad i introduced a regression while fixing the last issue you found. Now it should be good

One more item has to do with empty comments.

oyvuaw8xjy

bago2k4:
@thepug addressed

thepug:
Man, I hate that I am catching these one at a time but each change makes me try testing it differently.

An issue when changing the comment to blank while editing.

http://recordit.co/QZMyFNFBFG
qzmyfnfbfg

PR #326, opened 2018-01-09
Digest Settings

belucid:
https://trello.com/c/YBUH0SR6

This PR supports https://github.com/open-company/open-company-digest/pull/3

Depends on Auth - open-company/open-company-auth#23

- [ ] Review the code

To test

- [x] Ensure you're running w/ the latest Auth service (see above)
- [x] Change your users digest type and frequency, and TZ in the profile, does it "stick"?
- [ ] Merge
- [ ] Deploy to beta (if the rest of Digest isn't deployed yet, that's OK, just needs the Auth PR above)
- [ ] Trim your toenails


belucid:
NB: the "Slack" or "Email" digest picker is dumb. Ideally it would only show when we have the bot available to deliver the digest by Slack (see comment in the code).

Unfortunately, user profile is at `/profile` and we don't know what the team/org is. We're setting this digest setting for all the users' teams and orgs. This is a flaw in the backend data model. The digest settings should for the user per org, not just per user.

We'll need to move `/profile` under the org, such as `/acme/user-profile`. And we need to update the Auth data model for how we store digest settings. https://trello.com/c/W8Q5evKX

Can be this is a bit weird too because most the settings there are going to be for your whole user (name, avatar, password, email, TZ, etc.) but just a couple digest settings are different per org. We'll need visual design that helps make this clear. Maybe split global user settings and per org user settings into 2 different UIs?

PR #329, opened 2018-01-11
Signup auth refactor

bago2k4:
Another small piece of the app state refactor.
This is intended to move the user actions for sign up and auth to the `oc.web.actions.user` namespace. And the store changes actions to `oc.web.stores.user`.

To test:
- you need new Slack org that was never added to your auth DB (remove the auth DB if needed)
- [x] signup with the new slack org with adding the bot, everything good? Good
- [x] go to settings, can you add the Slack bot to the team? Good
- [x] go to invite, do you see the other Slack users? Good, try to invite one
- [x] go to share a post, can you share via Slack? Good
- [x] go to board settings, can you pick a slack channels for comments? Good
- [x] signup with email, everything good? Good
- [x] force a JWT refresh from console (read console help for instructions), did it work? Good
- [x] invite a new user via email (use an email you have access to), do you see the email (you need to start email service)? Good
- [x] click on the invite link in a new browser session, did it let you in? Good
- merge


src/oc/web/actions/user.cljs
```diff
@@ -60,4 +72,91 @@
...
+(defn signup-with-email []
+  (api/signup-with-email
+   (or (:firstname (:signup-with-email @dis/app-state)) "")
+   (or (:lastname (:signup-with-email @dis/app-state)) "")
+   (:email (:signup-with-email @dis/app-state))
+   (:pswd (:signup-with-email @dis/app-state))
+   signup-with-email-callback)
+  (dis/dispatch! [:signup-with-email]))
+
+(defn signup-with-email-data [signup-with-email-data]
```
thepug: I think we can clean this up.  The above function `signup-with-email` should accept `firstname` `lastname` `email` `pswd` as parameters.  We can get rid of the `signup-with-email-data` function.  The signup component can maintain the data instead of having it in the app state. On successful signup we can pass the user data to the data store if we still need to keep this data.

src/oc/web/stores/user.cljs
```diff
@@ -67,3 +71,53 @@
 (defmethod dispatcher/action :login-with-email/success
   [db [_]]
   (dissoc db show-login-overlay-key))
+
+;; Auth actions
+
+(defmethod dispatcher/action :bot-auth
```
thepug: This function is an action. It doesn't change the app state and uses functions unrelated to the app state, `router/get-token`, etc.  We should move this to user actions.  The current user action dispatches but doesn't need to and can just call this code.

bago2k4: Done

src/oc/web/actions/user.cljs
bago2k4: @thepug this is not possible because `signup-with-email` contains also the error states returned by the server. I agree this could be simplified but at this time it's beyond the scope of this refactor.
I changed the parameter name though because it was the same of the function name and could lead to problems.

thepug: Are the error states only from the server?  You can still use the error states but no need to have the other bits in the app state, `firstname`, etc.

bago2k4:
@thepug i tried to clean also the other references from user actions to app-state.

thepug:
With the latest changes I attempted to signup with slack and when I was done signing into slack I was kicked back to the index page instead of going through the NUX.

PR #330, opened 2018-01-11
Tablet breakpoint

bago2k4:
From this card:
https://trello.com/c/nKpWC0GP
Trying to address this item:
- Address above structural / breakpoint / reactive issues for tablet/narrow desktop, short desktop, and true mobile

There are still issues with activity and create post modals but it's not clear how they should behave on tablets.
For now this solves only these things on tablet window size:
- homepage
- about page
- slack page
- board view

### DIFF
--- a/scss/partials/_about.scss
+++ b/scss/partials/_about.scss
@@ -7,6 +7,12 @@ div#wrap {
     background: linear-gradient(to top, #FBFAF7 80%, #FBE3DD 85%, #FFEFBD 91%, #FFFBED 105%);
     margin-top: 1px;
 
+    section {
+      @include mobile_to_tablet() {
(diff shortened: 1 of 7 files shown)


scss/partials/_about.scss
```diff
@@ -141,7 +147,7 @@ div#wrap {
         max-width: 744px;
         margin: 16px auto 0px;
 
-        @include mobile(){
```
thepug: I'm not sure I understand the name change , `mobile` -> `tablet`.  Doesn't this css support both?

bago2k4: before it was only for mobile (<768px) now it's for tablet too (<~1200px)

PR #331, opened 2018-01-11
Board all caught up

bago2k4:
From this card:
https://trello.com/c/nKpWC0GP

Try to address this item:
- That's all for now message at the bottom of each board and AP


thepug:
Was this change suppose to have it only show the "all caught up" message on mobile, @bago2k4 ?

belucid:
Intent was only mobile @thepug and @bago2k4

thepug:
> Intent was only mobile @thepug and @bago2k4

Ok, @bago2k4 coded this to only show the message on mobile so I think that is correct.

PR #338, opened 2018-01-16
update to derivatives 0.3.0

martinklepsch:
I haven't tested this code but think it should work.
Derivatives 0.3.0 contains a breaking change which affects
the Om utilities in `rum-utils`. [Changelog](https://github.com/martinklepsch/derivatives/blob/master/CHANGES.md#030)


bago2k4:
Thanks @martinklepsch aside of a typo that i fixed looks like it works like a charm!

martinklepsch:
Oh haha 😄

PR #336, opened 2018-01-16
Mobile post edit

bago2k4:
Card: https://trello.com/c/nKpWC0GP

Improvements to the mobile editing. We decided to use always the entry-edit component to edit a post on mobile. Also this fixes the position of the alert modal.

To test:
- [x] create a new draft
- [x] create a new post (directly, no draft)
- [x] edit a draft
- [x] publish a draft
- [x] delete a post, do you see the confirmation dialog in the [right place](https://projects.invisionapp.com/share/BSF0L27RC#/screens/269910442)? Good
- [x] while you are editing try to scroll up and down, do you NOT see the board behind the edit view? Good
- [x] while you are looking at a post with reactions and comments (readonly) try to scroll up and down, do you NOT see the board behind the post view? Good

PR #340, opened 2018-01-17
Mobile share

bago2k4:
### Review after #336 

Card: https://trello.com/c/nKpWC0GP

To test on your mobile device:
- setup an org with slack
- make sure it has the bot
- load a board
- tap the vertical ellipsis
- tap on Share
- [x] share via slack
- [x] share via url (check the copy button actually copy the url)
- [x] share via email


belucid:
@bago2k4 

- [ ] the landing page for the share URL is not mobile ready yet

belucid:
@bago2k4 

- [ ] can we put metadata on the email field in the share dialog so we get the email keyboard treatment? Right now we get the keyboard w/ no `@` symbol and it wants to auto-upper case the first letter.

bago2k4:
@belucid i have changed the input type to be email so the mobile keyboard used is the right one. About the shared URL i think we should do a new PR for that, we can wait on this to merge or deploy but i won't do it here.
Do you know if a design exists? I can try guessing the best mobile layout with what we have and let Ryan react it. WYT?

bago2k4:
@belucid just pushed a new bug fix: if you share from a post view you don't see the share view until you dismiss the post view. Check also that it didn't regress on desktop because it touches that code too.

PR #341, opened 2018-01-17
Goodbye Om

bago2k4:
This branch is intended to lose our Om dependency. Until now we were using an om component as mount point that was bridging the app state derivatives to all our Rum components.
That was necessary because we had mixed Rum and Om components and we wanted to be able to use both.

At this point we have only 2 remaining Om components: org-dashboard and org-editing. I moved those 2 to Rum and remove the code that was bridging the app state to Om and then to Rum cursor (and to Martin's derivatives).

This is needed to simplify our code in preparation to use React 16. We can't upgrade to it right now because Rum is not ready yet. There is a PR open, still under discussion for some reason, but it will be ready soon.

To test build this branch and use the apps how you would normally do, try to use any feature you can think of.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -18,20 +18,17 @@
     ;; NB: Can't go past ClojureScript 1.9.908 right now due to issues w/ React
     [org.clojure/clojurescript "1.9.908"] ; ClojureScript compiler https://github.com/clojure/clojurescript
 
-    ;; Om and Rum React Frameworks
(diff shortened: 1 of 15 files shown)


belucid:
@bago2k4 can you deal w/ the merge conflict here? Don't want to gorp it up.

bago2k4:
@belucid ready to go

bago2k4:
@belucid solved conflict with mainline

PR #342, opened 2018-01-18
Mobile secure activity

bago2k4:
Card: https://trello.com/c/mvdlAHHA

To test:
- from desktop (or using the mobile-share branch if it's not merged already) share a post
- open it from your mobile device (in Private mode if you are logged in)
- [x] check that the view works on mobile
- [x] tap the footer Made with Carrot: does it NOT open the modal? Good (we don't have a mobile version just yet)
- [x] open the same on desktop and check that the same view didn't regress (check also the made with carrot button)
- [x] check with a very short post and with a very long post (better if it has images or some other rich content)

### DIFF
--- a/scss/partials/_made_with_carrot_modal.scss
+++ b/scss/partials/_made_with_carrot_modal.scss
@@ -54,10 +54,10 @@ div.made-with-carrot-modal-container {
       div.carrot-logo {
         position: relative;
         background-image: cdnUrl("/img/ML/carrot_logo.svg");
-        width: 82px;
+        width: 45px;
         height: 82px;
         background-repeat: no-repeat;
-        background-size: 82px 82px;
+        background-size: 45px 82px;
         background-position: center;
         margin: 0px auto;
       }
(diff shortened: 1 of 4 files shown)


belucid:
@bago2k4 

On mobile Chrome and FF the footer isn't fixed to the bottom. It's floated up. On Safari it's fine.

belucid:
@bago2k4 the footer graphic is stretched out oddly on desktop: http://take.ms/xGoiz

bago2k4:
@belucid i tried fixing it, i didn't fix it on all browser though because the fixed bottom bars have the problem of the resizing browser toolbars.
I was using the virtual height units for big web and mobile, it's a CSS unit that should always report the viewport height, problem is that the browser don't change this value when the toolbars appear/disappear, they say because
> Not only that looks like shit, but doing that at 60 FPS is practically impossible in most pages

More info here https://nicolas-hoizey.com/2015/02/viewport-height-is-taller-than-the-visible-part-of-the-document-in-some-mobile-browsers.html

So i moved to the js window height calculation and added a listener for the window RESIZE event. Now it's a bit more clear but not on all browsers the bar is always at the bottom, sometimes it has some space if you try scrolling the view.

Anyway try and let me know if we can leave with this.

The problem is only visible for small posts (smaller then the window height) and if the user scrolls, in the other cases it looks good in my tests.

I fixed also the stretched graphic.

PR #349, opened 2018-01-24
Independent topic

bago2k4:
Source: [QA doc]()

Item:
> You should be able to add/change topic name without going into Edit mode or needing to Save. It should be independent just as comments are independent.

Also this:
> Also notice what happens when you go into edit a topic. You see the last topic used on a diff card for a second before it clears out. https://www.useloom.com/share/37673cf34cc24001b32fb41e12f8e4a3

seems to be already solved, keep an eye open while you review this.

To test:
- open a board
- click New post
- add headline and body
- add a topic
- hit save as draft
- [x] did it save the draft with the correct topic? Good
- now edit the draft: change the topic
- post it
- [x] did it post with the correct topic? Good
- now expand the newly published post
- click on the topic
- remove the topic
- [x] do you see the post w/o the topic in the board? Good (refresh the board to make sure)
- expand the topic again
- click on + Add a topic
- add another topic
- save
- [x] did it save? Good

### DIFF
--- a/src/oc/web/components/activity_modal.cljs
+++ b/src/oc/web/components/activity_modal.cljs
@@ -229,6 +229,18 @@
         (reset! (::initial-headline s) initial-headline)
         (reset! (::edited-data-loaded s) true)))))
 
+(defn topic-did-change [s editing topic-map]
+  (let [activity-data (first (:rum/args s))]
+    (if editing
+      (do
+        (dis/dispatch! [:input [:modal-editing-data :topic-slug] (:slug topic-map)])
+        (dis/dispatch! [:input [:modal-editing-data :topic-name] (:name topic-map)])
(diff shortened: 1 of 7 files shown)


belucid:
@bago2k4 getting an odd thing where the reactions emoji appears in the topic dropdown (I also see it on staging right now):

http://take.ms/Pm7SA

src/oc/web/components/activity_modal.cljs
```diff
@@ -459,17 +471,21 @@
                                       :on-change #(close-clicked s nil)}))]))
               (when-not is-mobile?
                 (activity-attachments activity-data false))
-              (if editing
+              ; (if editing
```
belucid: do we need this commented code?

bago2k4:
@belucid fixed, good catches.

PR #350, opened 2018-01-24
Qa copy changes

bago2k4:
Source: [QA Doc](https://docs.google.com/document/d/1X3j869Y8n94zfqaGVhrWf8VOlr3qEa_4a1fxtIjH6fs/edit?pli=1#)

Items:
> Board delete prompt seems a bit too casual, “Delete this board?”, you’re really deleting all the content too. Could be dozens, hundreds of posts. Let’s use: “Are you sure? This will delete the board and ALL it’s posts, too.” as the copy.
 
> Change About Carrot link to “Support”; also when you click it the title should also change to “Support”

To test check the 2 copy changes.


belucid:
@bago2k4 

Let's update the copy to: "... and *all* it's posts ..."

And let's make it conditional on there being any posts. It's strange when the board is empty. In the case of an empty board, just use the first sentence.

PR #352, opened 2018-01-25
Big board perf

bago2k4:
Card: https://trello.com/c/qMwhbLQT

The current components tree is built like this:
```
                              org-dashboard
                            /        |      \
                           /         |       \
  entry-edit/activity-modal        navbar     dashboard-layout
                                                     |
                                                     |
                                               entries-layout
                                                     |
                                                     |
                                       activity-card * $number-of-posts
```
In the current mainline all these components are re-rendered when a change is made in the `entry-edit` component (or `activity-modal` in edit mode).
The change is made to the `:entry-editing` key of `oc.web.dispatcher/app-state`, this key is needed by `org-dashboard` to know if the editing is currently active (and show the edit component) but it's not needed by the sub-components.
With this fix the only component that updates when the user changes something in the edit component is `entry-edit` (or `activity-modal`), not the entire components tree.
The key point is this: https://github.com/open-company/open-company-web/pull/352/files#diff-8fee79a8f8336bc828df68f7992ab765R299
Instead of passing the whole entry-editing value to the org-dashboard i pass only a boolean value that says if it's editing (a new post or a draft in this case). This avoid the org-dashboard to re-render when a change is made inside the entry-editing value because the boolean value is still the same.

To test:
- use a board with 25+ posts (the more the better, you can use the [generate](https://github.com/open-company/open-company-storage#generate-sample-data) script of storage)
- load the board
- click New post
- try editing the headline and the body
- [x] do you NOT see a delay while you type? Good
- save to draft
- [x] are you redirected to drafts? Good
- [x] do you see the new post added to the drafts board? Good
- without refreshing the board
- click on the new draft
- edit the headline and the body
- [x] do you NOT see a delay while you type? Good
- now post it
- [x] are you redirected to the board where the post belongs? Good
- [x] do you see the post with the latest changes? Good
- now click on the post
- enter edit mode
- edit the headline and the body
- [x] do you NOT see a delay while you type? Good
- save the changes
- go back to the board and check you changes are visible
- merge


bago2k4:
[Post create/edit performance on larger (~20 posts) boards](https://trello.com/c/qMwhbLQT/1161-post-create-edit-performance-on-larger-20-posts-boards)

belucid:
Nice work! 👍

PR #353, opened 2018-01-30
Mobile search

thepug:
Changes to have the search look good on a small device.

Additional functionality to show search results when a user hits the back button or closes the clicked on result.

To test:

* open in a mobile device or mobile mode in your browser.
- [x] do you only see the spyglass?
* click on the spy glass
- [x] You should now see the X on the left and focus should be in the search input.
* search for something 
- [x] The screen should be white with "SEARCH RESULTS  No matching results"
- [x] when there are results there should be a scrollable list.
- [x] clicking on a result should open the post.
- [x] hitting the back button or the x will bring you back to the search results.
- [x] You can start a new search or clear the current search.

### DIFF
--- a/scss/partials/_navbar.scss
+++ b/scss/partials/_navbar.scss
@@ -102,9 +102,9 @@ nav.oc-navbar {
         font-size: 26px;
 
         @include mobile() {
-          width: 256px;
+          width: 200px;
           height: 32px;
-          margin-left: -128px;
+          margin-left: -107px;
           top: 0px;
         }
 
@@ -130,8 +130,10 @@ nav.oc-navbar {
       height: 24px;
 
       @include mobile() {
-        margin-top: 8px;
+        margin-top: 10px;
         margin-left: 16px;
+        width: 150px;
+        height: 40px;
       }
 
       button.mobile-navigation-sidebar-ham-bt {
@@ -144,7 +146,7 @@ nav.oc-navbar {
           background-repeat: no-repeat;
           background-size: 18px 14px;
           background-position: 3px 5px;
-          display: block;
+          display: inline;
         }
       }
     }
--- a/scss/partials/_navbar.scss
+++ b/scss/partials/_navbar.scss
@@ -102,9 +102,9 @@ nav.oc-navbar {
         font-size: 26px;
 
         @include mobile() {
-          width: 200px;
+          width: 40%;
           height: 32px;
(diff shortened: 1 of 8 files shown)


thepug:
[Mobile Search](https://trello.com/c/n31or2Nf/1134-mobile-search)

thepug:
scroll to bottom shouldn't show the board behind the results.  hide the board when search results are shown.

belucid:
SEARCH RESULTS (17) shouldn’t be fixed, should scroll away

thepug:
> SEARCH RESULTS (17) shouldn’t be fixed, should scroll away

So there was a reason I added that sticky header.  When scrolling the "triangle" will look odd if it scrolls pass an avatar.  I'll have to find a solution for this for this header to scroll away.

bago2k4:
@thepug i remember we discussed this together, the problem was only for desktop since there is the small arrow pointing to the search.
I guess you can use the fixed header on desktop that doesn't look so weird and not fixed on mobile.
Do you think this can work?

thepug:
> @thepug i remember we discussed this together, the problem was only for desktop since there is the small arrow pointing to the search.
I guess you can use the fixed header on desktop that doesn't look so weird and not fixed on mobile.
Do you think this can work?

Yeah that should work great.

thepug:
addressed feedback on the "SEARCH RESULTS" header and scrolling on mobile.

thepug:
Looks like the other items are broken
http://g.recordit.co/FvlVvXH1LQ.gif
fvlvvxh1lq

belucid:
@thepug 

2 issues here I think, first is the company logo and avatar logo are peaking out: http://take.ms/26NkA

The 2nd is that after tapping the spyglass it's pretty unclear what I'm supposed to do. "Search" there doesn't look like a text field I'm meant to tap on and start entering text into. I know we can't focus the field on mobile w/o a tap from the user... but this doesn't seem workable.

thepug:
https://opencompanyhq.slack.com/archives/C06SBQ3FD/p1518107281000806

on mobile put spyglass inside text input and change copy to "Enter your search"

thepug:
The issues mentioned here https://github.com/open-company/open-company-web/pull/353#issuecomment-364098942

are addressed.  The spyglass is now the input so you should always get the keyboard and it is clear you are able to search.

PR #357, opened 2018-01-31
Fix board scroll

bago2k4:
#### Just to make sure you can check all the test steps
___

This fixes a very annoying bug that was to reset the scroll position of the board when changing location to a post url and back to the board.
It fixes the problem avoiding to have the whole components tree reloaded because it avoids to trigger a url change. It replaces the url with a pushState and replace the internal app router so the components show the post modal.

To test:
- open your browser window enough 
- open a board
- scroll down
- click on a card
- [x] does the post modal appear? Good
- [x] do you NOT see the board changing above and below the post modal while it appeared? Good
- dismiss the modal
- [x] are you back to the board at the same scroll position? Good
- open a digest link
- [x] do you see the post modal correctly populated? Good
- dismiss the post modal
- [x] did you NOT see a spinner in the transition from the modal to AP? Good
- merge


README.md
```diff
@@ -195,6 +195,8 @@ provided in the table below:
 |  :activity-delete/finish | API | Entry delete request finished, refresh the board data. |
 |  :activity-get/finish | API | Activity (secure or not) loaded from API. |
 |  :activity-modal-edit | UI | Activate editing in modal. |
+|  :activity-modal-fade-in | UI | Show the activity modal without reloading the route. |
```
thepug: I'm not sure why this was moved to a dispatch when it is an action creator that ends up using a set route.  I don't see a need for this to be part of the application state since I can't see where any other component uses it.

I suggest moving this code to a new file `/actions/[activity|entry|post].cljs`  Not sure what the naming convention is?

src/oc/web/actions.cljs
```diff
@@ -1114,12 +1114,48 @@
...
+    (.pushState (.-history js/window) #js {} "" to-url)
+    (router/set-route! [org to-board (if from-all-posts "all-posts" "dashboard")]
+     {:org org
+      :board to-board
+      :activity nil
+      :query-params (:query-params @router/path)
+      :from-all-posts false}))
+  (-> db
+    (dissoc :activity-modal-fade-in)
+    (dissoc :dismiss-modal-on-editing-stop)))
```
thepug: I believe this isn't used anymore.

src/oc/web/actions/activity.cljs
```diff
@@ -0,0 +1,125 @@
...
+      :board to-board
+      :activity nil
+      :query-params (:query-params @router/path)
+      :from-all-posts false}))
+  (dis/dispatch! [:activity-modal-fade-out board-slug board-filters]))
+
+(defn entry-edit
+  [initial-entry-data]
+  (load-cached-item initial-entry-data :entry-editing)
+  (dis/dispatch! [:entry-edit initial-entry-data]))
```
thepug: I can't see anything that handles this dispatch. I don't see anything that does in mainline either.

screenshot

thepug:
Issue with the text jumping.

http://g.recordit.co/UjEeUcIfmc.gif
ujeeucifmc

bago2k4:
@thepug should be good now..

thepug:
An issue with the saved data while editing.  Its tough to take a movie of this so I'll try to describe the steps.

* Create a new post
* add data and close out the browser
* You will be prompted if you want to leave and choose leave.
* open the site back up and click new post again.
* the saved data is property used
* click save draft
* once the draft is saved click on all posts
* scroll down to the second to last item
* click on the edit button.    I don't see the post's content instead I see the previously saved data.

bago2k4:
@thepug i can't replicate this.. i tried with the console open and the indexed DB record of the not saved post is removed when i save as draft. I tried with the console closed too and the result is the same, the post i see is the correct one.
Did you try rebuilding the code when you switch to this branch? Sometimes i have problems if i don't rebuild from scratch.

thepug:
Here is a version of the issue:
http://g.recordit.co/jrcTQxKPZ9.gif

jrctqxkpz9

bago2k4:
@thepug great catches! now it should be solved... this bug was around for some time, we had this weird body appearing every now and then while entering editing.. now we know why!!

PR #358, opened 2018-01-31
Refactor user actions for the initial API requests.

thepug:
Cleanup for the API entry handlers.

To test:
- [x] signup with slack - new team - ensure that this goes through NUX ,etc

- [x] login with slack 

- [x] signup with email - new team

- [x] login with email

- [x] click the verify email link and ensure it logs the user in and verifies the user.

- [x] invite email - goes through the NUX etc and verifies the user.

- [x] reset password - can you reset your password?


src/oc/web/actions/user.cljs
```diff
@@ -42,8 +43,30 @@
 (defn jwt-refresh []
   (api/jwt-refresh update-jwt logout))
 
-;; Login
+;; API Entry point
+(defun entry-point-get-finished
```
bago2k4: @thepug You don't need `defun` for this signature, you can use `defn` with this method definitions.

src/oc/web/api.cljs
```diff
@@ -4,6 +4,7 @@
   (:require [cljs.core.async :as async :refer (<!)]
             [cljs-http.client :as http]
             [defun.core :refer (defun-)]
+            [defun.core :refer (defun)]
```
bago2k4: @thepug why did you use 2 different require?
you can do
```
   [defun.core :refer (defun- defun)]
```

src/oc/web/api.cljs
```diff
@@ -258,12 +259,15 @@
         (fn [{:keys [success body status]}]
           (dispatcher/dispatch! [:org (json->cljs body) true]))))))
 
-(defn get-auth-settings []
-  (auth-http http/get "/"
-    {:headers (headers-for-link {:access-control-allow-headers nil :content-type "application/json"})}
-    (fn [response]
-      (let [body (if (:success response) (:body response) false)]
-        (dispatcher/dispatch! [:auth-settings body])))))
+(defun get-auth-settings
```
bago2k4: @thepug Same as above, don't need `defun` for this, just use `defn`.

src/oc/web/core.cljs
```diff
@@ -115,7 +115,17 @@
       (when (or (> (- now latest-entry-point) reload-time)
                 (and (router/current-org-slug)
                      (nil? (dis/org-data))))
-        (api/get-entry-point))
+        (api/get-entry-point
```
bago2k4: @thepug we can use `user-actions/get-entry-point` here, right?
we are using action creators to avoid having api calls and similar things around the code.
WYT?

src/oc/web/core.cljs
```diff
@@ -150,13 +160,6 @@
     (router/set-route! (vec (remove nil? [route-name org])) {:org org :query-params (:query-params params)}))
   (post-routing)
   (when-not (contains? (:query-params params) :jwt)
-    (when (contains? (:query-params params) :login-redirect)
-      (cook/set-cookie!
-       :login-redirect
-       (:login-redirect (:query-params params))
-       (* 60 60)
-       "/"
-       ls/jwt-cookie-domain ls/jwt-cookie-secure))
```
bago2k4: @thepug you removed all this code, did you replace it with something else? We don't use much the login redirect but in some places i think it's still necessary..

thepug: So I went through mainline and tested all the places I could see login redirect being used and none of it was used. The app just redirects me to the current board after logging in.  I removed all this code because of that.

PR #359, opened 2018-02-01
Improve mobile edit

bago2k4:
Design: https://projects.invisionapp.com/share/VAFLD1KR9BG#/screens

To test:
- on your mobile device
- tap new post
- tap Post (w/o title and body)
- [x] do you see the tooltip suggesting to add at least a title? Good
- add a title
- Save as draft
- [x] are you redirected to the Drafts board? Good
- [x] do you see the new draft? Good
- select the draft
- add a body
- post it
- [x] are you redirected to the board? Good
- [x] do you see the new post? Good
- tap new Post again
- add a title and a body
- tap on the board name
- switch board
- Post
- [x] are you redirected to the correct board? Good
- [x] do you see the new post? Good
- tap on the vertical ellpsis of the FoC
- tap on edit
- edit the post (best the title so it's visible from the FoC)
- Save
- [x] do you see the changes from the board? Good
- merge


belucid:
@bago2k4 

- [ ] after tapping new post, I see the placeholder for body but no placeholder for title. Even though title doesn't have focus and my keyboard isn't popped up.
- [ ] "Untitled Post" (I can get it after making the body get focus doesn't seem like a placeholder.
- [ ] When body gets focused it scrolls so the body starts at the very top. I find this disorienting and not necessary. User can handle their own scrolling as they want to in the body.
- [ ] Successful save should return you to the board. Seems like it didn't work when you stay in place. (and besides, the sequence Edit -> Save -> X gives you the unsaved changes prompt, which isn't good, but doesn't need fixed if we close on save)

bago2k4:
@belucid addressed he above items, the first 2 will be changed with the new design so i did them but w/o losing much time on them.
The third is automatic iOS behaviour, i resized the minimum body field height because it tries to scroll the field to be all visible when you focus on it, so now it does what you see in the attached image.
The fourth was never an issue for me, i probably fixed it in another branch.. let me know if you still see it.
img_0510

PR #361, opened 2018-02-02
Fix error pages redirects

bago2k4:
BUG: Right now when you land on one of our error pages (404 or 5xx) you are redirected to /404?path=/path/that/404ed (same for /500) then the URL is rewritten with pushState to /path/that/404ed. This is because we always serve the /app-shell.html from nginx so we can't change the status from there. The problem is that to go back to where you came from you need to click 2 times the back button because your history look like this:
- www.where.com/you/came/from/here
- carro.io/404?path=/path/that/404ed
- carrot.io/path/that/404ed

if we use replaceState instead of pushState you have this final history:
- www.where.com/you/came/from/here
- carrot.io/path/that/404ed (replaces carro.io/404?path=/path/that/404ed)

and with one single back click you can go back to the initial page.

To test try to visit a not existing page: `/this-doesnt-exists` for example, if that works the 500 will work just the same since it uses the same code.
If you are being scrupulous today insert `(router/redirect-500!)` in a route of `oc.web.core` and visit a path that match that route.
In both cases make sure you see the correct page (404 or 500) and that if you click back you go back to the initial page not to `/404?path=...` or `/500?path=...`.


belucid:
Test didn't work @bago2k4 

Steps I used:

- merged into `merge-branch`
- deployed to staging
- visited: staging /
- visited: /no-existence
- got redirect to /no-existence as 404 page
- hit back button in browser
- went back to redirect and ended up back on /no-existence as 404

Browser is FF.

bago2k4:
You are right @belucid i fixed only the second redirect, but you need to go back 2 pages:
google.com ->> /non-existing -> /404?path=/non-existing -> /non-existing
->> actual history navigation
->   replace current history object so the back button is not affected

Now you should be able to go back to the initial location

PR #366, opened 2018-02-07
Web app version

bago2k4:
Card: https://trello.com/c/R3D7KXoF

### Review with: https://github.com/belucid/open-company-infrastructure/pull/55
___

This adds a static file when the build finishes located at `/version/version{ deploy-key }.json` (from web, physically it's at `{ oc_web_project_root}/target/public/version/`).
The client loads it every time the app loads and every now and then: it uses the same logic used to refresh the auth and storage entry point, it's cached for 1 hour top.

If the file 404s it means we are using an old version of the app because the json file was removed and another one with the new deploy-key was added.
If it gets a 200 or a 304 (cached by nginx) it means it's using the latest code available.

When it finds out the code is old it shows our usual red banner prompting the user to refresh. This UI will be refreshed soon.

We initially thought to use ETAG and the last modified headers of the page but the response headers aren't available through javascript, to achieve it should load the app shell another time via ajax to get those values so i preferred to add a new simple json file that is lighter and faster to load.

As said this is not a real time check, to do that we should use our change service web socket but for this scope it's enough this way at the moment

NB: i added a static /version/version.json file used when developing, i didn't remove the call because it's useful to see it in the console to make sure it's happening even if it's not needed.

To test:
- change the entry point reload time to 60 seconds `oc.web.core:118` to `reload-time (* 60 1000)]`
- deploy the branch with this new change
- open a board page on staging (or the env deployed)
- switch to network tab, filter for XHR and search for `version`
- don't close this window
- now deploy again the code (it will take more than one minute, if not make sure 1 minute is passed)
- go back to the opened board window
- click on another board or on all-posts
- [x] do you see another GET for `/version/version{ deploy-key }.json`? Good
- [x] did it 404? Good
- [x] do you see the banner saying you need to refresh? Good
- **rollback the commit**
- merge


bago2k4:
[Check for out of date app and prompt user to refresh](https://trello.com/c/R3D7KXoF/1101-check-for-out-of-date-app-and-prompt-user-to-refresh)

PR #373, opened 2018-02-19
Needed to support private board notifications. 

thepug:
When sending new board data also send user information that can be used to send notifications (slack or email).

See instructions to test at https://github.com/open-company/open-company-auth/pull/28


belucid:
@thepug 

Here's what I'm seeing during some Slack testing of this:

Have 2 users, 1 email and 1 Slack. We'll call the uS and uE. (uS also  happens to be an admin, and uE happens to be a contributor, could be that's what's causing the difference, not Slack vs. Email).

Go to create a private board with uS. As soon as I toggle to private, uS is shown as a "Full Access" member uE us in the dropdown list (so all is as expected).

Go to create a private board with uE. As soon as I toggle to private, uE is NOT shown as a member and uS is in the dropdown list. If I pick uS from the dropdown list however, it does not show up in the member list, and the dropdown disappears.

So I can't really test the Slack part of this with this web branch.

thepug:
replying to https://github.com/open-company/open-company-web/pull/373#issuecomment-369233244

@belucid I dont' see this behavior.  Maybe the email user isn't verified?

belucid:
@thepug so where I ended up with in testing this @thepug is, if a Slack user invites an email user, they get an email, but if an email user invites a Slack user, they get an email, not a Slack.

PR #384, opened 2018-02-26
Only render emoji-picker on click

thepug:
The quick fix for the performance issues with rendering the emoji picker.


thepug:
[Performance improvements for All Posts](https://trello.com/c/WrlxSJlg/1180-performance-improvements-for-all-posts)

src/oc/web/components/ui/emoji_picker.cljs
```diff
@@ -136,23 +136,24 @@
...
-                           (.focus @last-active-element))
-                         (reset! visible false)
-                         (when (fn? add-emoji-cb)
-                           (add-emoji-cb @last-active-element emoji add-emoji?))))}))]]))
\ No newline at end of file
+     (when @visible
+       [:div.picker-container
+         {:class (utils/class-set {position true
+                                   :preloading (not @(::preloaded s))
+                                   :visible @visible})}
```
bago2k4: I think we can remove this `visible` class since we hide the picker w/o rendering it at all now. Before it was always rendered inside the page even if it was hidden for most of the time.

bago2k4: Also ::preloading is not used anymore since it only renders when it's needed, we can remove all that code too.

PR #389, opened 2018-02-28
Don't ignore the same user's comment updates.  

thepug:
Previously we would ignore a web socket comment update if from the same user.  This change no longer ignores it and only updates the body and updated-at fields.

To test:
open a browser with one user
open a browser with the same user.
add a comment to a post
- [x] When you edit a comment does it show up the same in both browsers?
- [x] Can you edit again in the different browser?

- [x] merge


thepug:
@belucid are you still having issues with this one?  You were having issues where it wouldn't watch a board on connection?

PR #390, opened 2018-03-01
Render error catch

bago2k4:
With React 16 we have the possibility to provide a different DOM if any of our components is throwing an error inside the render cycle.

This is a quick try to see how good it would be to redirect to our 500 page when a component errors.

Probably we need to provide a better page for this, 500 is internal server error, this is more internal client error. Let's discuss.

To test insert this statement `(throw (js/Error. "Oops"))` inside the render cycle of a component, then deploy to staging and visit the path that uses the above component.
Then check sentry to see if the error is there and if the information is enough to fix the problem.
Staging has already this branch but without the throwing error line.
You can't use localhost because it would be missing some information, here is an error thrown with my local sentry: https://sentry.io/opencompany/oc-iacopo-web/issues/476764005/


bago2k4:
Closing for now, added this to backlog https://trello.com/c/ZzBzrkLS

PR #395, opened 2018-03-02
Watch boards that are in all posts. 

thepug:
Unwatch all boards and send a single watch each time a board change occurs.

In addition to the watch board changes I added some code to "listen" to dispatches in the reactions store. The code simply reads the app state data and builds a hashmap with the uuid and app state key. This hashmap will then be used to lookup the keys when data arrives over the websocket.  I only changed this for comment reactions.  The other parts make an API request to get the latest data.
It is best to view the reaction store in full view because this diff is a bit crazy. 
https://github.com/open-company/open-company-web/pull/395/files#diff-5af6f3ec0cecdb163d32fa028d467d3e

Needs the following PRs to work:
https://github.com/open-company/open-company-interaction/pull/17
https://github.com/open-company/open-company-storage/pull/113
https://github.com/open-company/open-company-lib/pull/27

To test you will need two users logged in to the same team.

- Open up all posts as user one.
- Open up the first post shown in user one's all post list as user org. (can also be done in all posts view)
- As user two make a comment on the post.
- [x] Does the comment show up in user one's all posts view?
- [x] As user one can you react to the comment by clicking the thumb?
- As user two react to the post
- [x] Does the reaction show in user one's all posts view?
- As user two edit the comment created.
- [ ] Does the edit show up in user one's all posts view?
- As user two delete the comment.
- [ ] Does the comment get removed from user one's all posts view?


thepug:
[Real-time Interactions for All Posts](https://trello.com/c/LXoyltST/1199-real-time-interactions-for-all-posts)

bago2k4:
@belucid i found an issue when reacting to an entry with another user. The app-state wasn't updated when the reaction-add came in.
With the last commit i fixed the issue. Now the entry reaction uses the same system of the comment reaction to look and update the related data in the app-state.
I think you should take a look or/and @thepug and test again this.

belucid:
OK @bago2k4 I'll test again with the new commit. Thanks.

belucid:
Stopping short of testing edit/delete of comments for now since they aren't in the UI, this will get tested thoroughly when we add them back.

belucid:
@thepug and @bago2k4 merging this one with the race condition between Auth'ing a new Interaction service web socket connection and watching the board(s) since this isn't a new issue to this branch, just came up while testing this branch.

Resolving this has been queued up for after the refactor.

PR #402, opened 2018-03-08
New sections UI

bago2k4:
This is the editing part of this card: https://trello.com/c/0sIPXuJH

It changes the old board settings component with the new section editor. Also changes the old list of boards to move a post with the new sections picker where the user can create a brand new section without leaving the post editing.


bago2k4:
[Sections](https://trello.com/c/0sIPXuJH/1207-sections)

PR #406, opened 2018-03-13
Changes to the technical documentation.

thepug:
This outlines the design of our flux pattern and describes where the action creators are, what stores we use and the app state we use to trigger render events in our components.


docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+
+- https://github.com/facebook/flux - The Flux repo (documentation of the pattern, as well as a JavaScript Dispatcher implemntation)
+- https://github.com/facebook/flux/tree/master/examples/flux-concepts - High-level Flux concepts
+- https://facebook.github.io/flux/docs/in-depth-overview.html - Flux in depth
+- https://github.com/kgann/cljs-flux - An example implementation of the Flux pattern in ClojureScript
+
+### Action Creators
+
+An action creator is called by a user interaction or an API request finishing.
+
```
belucid: I think a key piece of info here @thepug is what our action creators do. Flux leaves this very abstract, just that in the end they emit an action to the Dispatcher that has the info needed about what state changed, and that the action be semantic, not overly impl. specific.

But what's interesting to us is, what does it mean to us about what an action creator will do. The responsibilities I see are things like:

- Pass an action to the dispatcher
- Make API invocations
- Update cookies
- Update URL routes

Do each of these belong in an action creator? Is this exhaustive? Are there other responsibilities?

docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+These action creators are located in `/src/oc/web/actions/`. They are losely
+organized based the data they will act on. So `/src/oc/web/actions/user.cljs`
+contains actions for a user to sign up, login, and change user profile
+information.
+
+
+### Simplified Data Model
+
+- Org
+  - has many Teams
```
belucid: @thepug this seems backwards to me. Our auth model is that a team has many orgs, not vice versa.

docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Action Creators
+
+An action creator is called by a user interaction or an API request finishing.
+
+These action creators are located in `/src/oc/web/actions/`. They are losely
+organized based the data they will act on. So `/src/oc/web/actions/user.cljs`
+contains actions for a user to sign up, login, and change user profile
+information.
+
```
belucid: @thepug IMHO it would be worthwhile to have a code snippet of a typical action creator pair... maybe `query` and `query-finished` or `login-with-email` and `login-with-email-finish`.

docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+  - has an author
+
+- Comment
+  - has many Reactions
+  - has an author
+  
+### Stores (changing the app state)
+
+Our action creators will dispatch data and an event using our dispatcher.
+We subscribe stores to particular events by creating a function with an arity
```
belucid: @thepug This doesn't match Flux, yeah? Flux says all stores get all events and decide if they care about it rather than subscribing to particular events.

docs/TECHNICAL_DESIGN.md
belucid: Alternatively, we could just make up examples. Maybe a simple action that doesn't need API and one that's paired with a finish.

docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+matching the event and payload. Stores with interest in the same event/payload
+will have a separate register to the dispatcher.
+
+When parts of the app state are changed a derivative is changed and used to
+render the UI. Also UI components can read from a store to retrieve more information.
+
```
belucid: @thepug Similarly I think it would be good to show code of a simple store or an example. In particular, should be clear: registering with the dispatcher, receiving an action, and changing a derivative.

docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,155 @@
...
+  - has an author
+  
+### Stores (changing the app state)
+
+Our action creators will dispatch data and an event using our dispatcher.
+We subscribe stores to particular events by creating a function with an arity
+matching the event and payload. Stores with interest in the same event/payload
+will have a separate register to the dispatcher.
+
+When parts of the app state are changed a derivative is changed and used to
```
belucid: Seems our use of derivatives needs some explanation here. Base Flux talks about stores holding the state of an app and emitting a change event. It's not trivial to map this onto what we're doing with app state and derivatives, so, need more info here.

docs/TECHNICAL_DESIGN.md
```diff
@@ -0,0 +1,153 @@
...
+
+#### Comment
+
+Information about comments.
+
+#### Search
+
+Handles search results.
+
+### App State Design
```
belucid: @thepug this section seems out of date now, no?

docs/TECHNICAL_DESIGN.md
thepug: This is different.  I added more documentation in hopes to clear this up.

docs/TECHNICAL_DESIGN.md
thepug: I simplified the example. It is only meant to show that it is a single atom and we use "keys" in the atom to specify parts of the atom that we are interested in.  Updated documentation in hopes to clear this up.

PR #434, opened 2018-03-26
REFACTOR: Use core.async pub/sub for the websocket clients.

thepug:
This changes the web socket interaction and change client. The change uses `core.async`'s publish / subscribe features to support one to many event handlers.

The subscribe functions are a bit strange because of the namespace issues in `actions.cljs`.  These will be cleaned up during the refactor changes.

To test:

Login with different users A/B
- As user B, open up a single section url.
- As user A , add a new post to the section above.
- [x] Does the new post show up in user B's section list?
- As user A, add a comment to the post.
- [x] Does the comment show up in user B's post?
- As user A, add a reaction to the post.
- [x] Does the reaction show up in user B's post?
- As user A, remove the same reaction from the post.
- [x] Does it get removed in user B's post?
- As user A, react to the comment.
- [x] Does it the reaction show on the comment in user B's display?
- As user A, remove the reaction to the comment.
- [x] Does it remove the reaction on the comment in user B's display?
- As user A, edit the comment.
- [ ] Does the update show in user B's display?
- As user A, remove the comment.
- [ ] Does the comment get removed in user B's display?

### DIFF
--- a/src/oc/web/actions.cljs
+++ b/src/oc/web/actions.cljs
@@ -11,6 +11,8 @@
             [oc.web.actions.user :as ua]
             [oc.web.actions.team :as ta]
             [oc.web.actions.activity :as aa]
+            [oc.web.actions.comment :as ca]
+            [oc.web.actions.reaction :as ra]
             [oc.web.lib.jwt :as jwt]
             [oc.web.lib.utils :as utils]
             [oc.web.lib.cookies :as cook]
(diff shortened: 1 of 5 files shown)


belucid:
Can't test the final 2, but those will be tested well when we add back in comment edit/delete.

bago2k4:
@belucid i went through the last 2 steps and they worked just fine, checking them out

PR #441, opened 2018-03-28
More cleanup

bago2k4:
Card: https://trello.com/c/ovJDfh65

Development build has a 404 since the app-shell was still including svg-utils.js file that was removed.
Also this removes the carrot-tip component (also called bat-signals) since we don't have that kind of NUX anymore.

To test:
- signup
- pass NUX
- add a post
- edit a post
- add a section
- [x] everything worked? Good


bago2k4:
[More code cleanup](https://trello.com/c/ovJDfh65/1244-more-code-cleanup)

PR #447, opened 2018-03-30
Emoji comments

bago2k4:
Card: 
Item:
> We no longer have the special treatment for a comment that is only emoji working (should be shown bigger with no agree action)

We use `oc.web.utils.comment/is-emoji` function to check if it's an emoji comment but it's not perfect. It doesn't recognise all the emojis but at least it's safe in the sense that it should not show normal comments as emoji comments.

To test:
- go to stream view
- add a normal comment
- [x] do you see it? Good
- [x] can you agree to it? Good
- [x] font size of the body is 14px? Good
- now add a single emoji comment
- [x] do you see it? Good
- [x] can you NOT react to it? Good
- [x] font size of the emoji is 50px? Good
- retry with some of the most common emoijs

NB: it should also recognise up to 3 emojis (maybe more) since some emojis are composed with up to 11 code points so it's hard to distinguish them with a simple reg exp.

### DIFF
--- a/scss/partials/_stream_comments.scss
+++ b/scss/partials/_stream_comments.scss
@@ -118,6 +118,16 @@ div.stream-comments {
           &>*:last-child {
             margin-bottom: 0px;
           }
+
+          &.emoji-comment {
+            font-size: 50px;
+            height: 50px;
+
+            // & > * {
+            //   line-height: 20px;
+            //   margin-top: 20px;
+            // }
+          }
         }
       }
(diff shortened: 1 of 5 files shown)


bago2k4:
@thepug fixed

thepug:
@bago2k4 `is-emoji` is returning nil for all the tests I have made.

bago2k4:
@thepug fixed

PR #448, opened 2018-03-30
Fix user email update

bago2k4:

Item:
> Changing email address of user in user profile doesn't work (may be backend, frontend or both)

To test:
- go to /profile
- change your email
- Save
- [x] do you see the new email still in place when save finished? Good
- refresh the page
- [x] do you see still the new email? Good
- [x] repeat changing first name (or the last name) and email

### DIFF
--- a/src/oc/web/components/user_profile.cljs
+++ b/src/oc/web/components/user_profile.cljs
@@ -194,7 +194,7 @@
                   {:type "text"
                    :tab-index 1
                    :on-change #(change! s :first-name (.. % -target -value))
-                   :value (:first-name current-user-data)}]]]
+                   :value (or (:first-name current-user-data) "")}]]]
             ; Current password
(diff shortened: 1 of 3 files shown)


src/oc/web/components/user_profile.cljs
```diff
@@ -194,7 +194,7 @@
                   {:type "text"
                    :tab-index 1
                    :on-change #(change! s :first-name (.. % -target -value))
-                   :value (:first-name current-user-data)}]]]
+                   :value (or (:first-name current-user-data) "")}]]]
```
thepug: This cleanup should be done once in the user store.  Before changing the `:current-user-data` check to see if the `:first-name` is nil and add the empty string.  Then change the app state. You only do it once instead of all the times you change it here.

bago2k4:
@thepug fixed

PR #449, opened 2018-03-30
Fix user profile timezone

bago2k4:

Item:
> Brand new user, go to user profile, click dropdown and change digest type from Email to Slack, at the same time, you'll see Time Zone changes from US/Eastern to America/New York. Why does this happen?

To test:
- go to /profile
- [ ] do you see the timezone? Good
- [ ] is it the same of your browser timezone? Good
- switch from slack to email or the opposite
- [ ] do you NOT see the timezone change? Good

### DIFF
--- a/src/oc/web/components/user_profile.cljs
+++ b/src/oc/web/components/user_profile.cljs
@@ -118,6 +118,14 @@
       :else
       (user-actions/user-profile-save current-user-data edit-user-profile))))
 
+(defn- setup-timezone [s]
+  (when (and (not (utils/is-test-env?))
+             (empty? (:timezone (:user-data @(drv/get-ref s :edit-user-profile)))))
+    (dis/dispatch!
+      [:input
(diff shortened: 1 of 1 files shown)


src/oc/web/components/user_profile.cljs
```diff
@@ -133,17 +141,13 @@
                           (rum/local false ::current-password-error)
                           {:will-mount (fn [s]
                             (dis/dispatch! [:user-profile-reset])
+                            (setup-timezone s)
```
thepug: This is also a sign of code that should exist in the user store.  This case is a bit different because it currently uses `:input` but should probably be changed to include a timezone in the user store. As always it is possible we need the user's timezone in other components.

bago2k4:
@thepug fixed this too in #448 since it was data to setup in the store, you can close this when you verify test steps are working there

PR #455, opened 2018-04-02
Remove unused keys from dispatcher data.

thepug:
These data keys are no longer used so removing.


thepug:
[Remove unused app state items.](https://trello.com/c/xxNF7kTn/1242-remove-unused-app-state-items)

bago2k4:
i removed some more @thepug, some derivatives and a couple of not used functions

PR #456, opened 2018-04-03
If at least one of the teams has Slack enabled let the user select it.

bago2k4:

Item:
> Can pick digest type as Slack in User Profile even though there's no bot to send the Slack digest

Assigning this to you @belucid because i think you found the bug.

I changed it so now it checks if at least one of the teams the user is associated to has a slack bot. If so i guess the backend will use the available bot for that team and email digest for the other. Makes sense?
Do you think this should be associated with a backend PR?

### DIFF
--- a/src/oc/web/components/user_profile.cljs
+++ b/src/oc/web/components/user_profile.cljs
@@ -1,6 +1,7 @@
 (ns oc.web.components.user-profile
   (:require [rum.core :as rum]
             [org.martinklepsch.derivatives :as drv]
+            [oc.web.lib.jwt :as jwt]
             [oc.web.urls :as oc-urls]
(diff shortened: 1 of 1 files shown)


belucid:
@bago2k4 I do think there is hole here where 1 team has the bot and another doesn't, but we aren't really dealing much w/ multi-team users in the real world yet and we already have a tech debt card for this problem: https://trello.com/c/W8Q5evKX

So leaving it at that for now.

belucid:
hmm don't think this is really resolved yet @bago2k4 

We can still have an email only user on a Slack team w/ the bot, and they are able to select Slack as their digest. We can discuss more what to do here.

PR #457, opened 2018-04-03
Fix duplicate tooltips

bago2k4:

Item:
> Duplicate tooltips when hovering date and have conflicting info (the grey text one seems accurate): http://take.ms/Ro82q

To avoid the browser showing 2 times the tooltip (one with bootstrap style and the other with the native tooltip style) i switched to use data-title as attribute, bootstrap read it just like the usual `title`.
Since the browser reads only the `title` attribute it shouldn't be possible to see both tooltips together.

Assigning this to you @belucid because i think  you found it and are able to repeat test steps


belucid:
Needs a backend fix in storage re: to a board w/ Slack posting enabled.

PR #459, opened 2018-04-03
Fix edited label

bago2k4:
Item:
> Do not use :updated-at for the edited but use the list of authors to see if it was always edited.

BUG: after creating a new post i was seeing a label reporting an edited at already. I checked and the web was using the wrong attribute to check when the post was edited.
This PR fixed this, but also needs a backend related PR to avoid getting 2 record in the author list right after creation (not on initial creation, but on a second GET right after).

Assigning this to @belucid because we discussed in Slack and he's also working on a backend fix.


belucid:
@bago2k4 please resolve the conflict here and I'll get back to working on this one.

bago2k4:
@belucid this is good again

bago2k4:
@belucid fixed conflicts here too

bago2k4:
@belucid i was looking again at this PR and i think we don't need anymore.

This PR was meant to always use :published-at to display the date of a post, but now we use the same components for drafts board. This means sometimes a post can have only :created-at. Using the conditional `(or (:published-at post) (:created-at post))` is ok.

Also this was setting the time since in each post but i think it's better to re-calculate it on each render so the time is updated more often.

PR #460, opened 2018-04-03
Fixes the issue of the Save button being stuck as 'Saving....'

thepug:
To test open up team settings and change the team name.  
- [x] Does the button move from 'Saving...' to 'Saved!'?


thepug:
[Bug when saving team info.](https://trello.com/c/XjISUOAD/1257-bug-when-saving-team-info)

PR #479, opened 2018-04-06
REFACTOR: Cleanup stores that have oc.web.router dependencies.

thepug:
To test this one you will need to do the following and match the expected behavior.

You should also have a second user open viewing the changes and verify they occur in both clients.
- [x] Create a new post
- [x] Delete a new post
- [x] Open a post in full screen mode. 
- [x] Create a new section 
- [x] Switch to a different section
- [x] Add a reaction to a post
- [x] remove a reaction to a post
- [x] add a comment to a post
- [x] react to a comment
- [x] remove reaction to a comment


thepug:
@bago2k4 updated to fix the comment add over the web socket

src/oc/web/actions/activity.cljs
```diff
@@ -300,10 +299,15 @@
...
 (defn get-entry [entry-data]
-  (api/get-entry entry-data
-    (fn [{:keys [status success body]}]
-      (if success
-        (dis/dispatch! [:entry (:uuid entry-data) (clj->js body)])))))
+  (let [is-all-posts (or (:from-all-posts @router/path)
+                         (= (router/current-board-slug) "all-posts"))
+        board-key (if is-all-posts
+                   (dis/all-posts-key (router/current-org-slug))
+                   (dis/board-data-key (router/current-org-slug) (router/current-board-slug)))]
```
bago2k4: @thepug you can avoid all this `let` using `dispatcher/current-board-key` that does the same as you do here

src/oc/web/actions/activity.cljs
```diff
@@ -385,8 +389,12 @@
               (oc-urls/org org-slug))))))))
 
 (defn activity-delete [activity-data]
-  (api/delete-activity activity-data activity-delete-finish)
-  (dis/dispatch! [:activity-delete activity-data]))
+  (let [is-all-posts (or (:from-all-posts @router/path) (= (router/current-board-slug) "all-posts"))
+        board-key (if is-all-posts
+                    (dis/all-posts-key (router/current-org-slug))
+                    (dis/board-data-key (router/current-org-slug) (router/current-board-slug)))]
```
bago2k4: @thepug same as above

src/oc/web/actions/activity.cljs
thepug: Yeah, the old cut - n - paste from old code.  Changing!

thepug: pushed the changes

thepug:
updated to address https://github.com/open-company/open-company-web/pull/479#pullrequestreview-110965222

src/oc/web/stores/comment.cljs
```diff
@@ -245,6 +244,7 @@
             sorted-comments-data (sort-comments new-comments-data)
             org-slug (name (nth entry-key 0))
             board-slug (name (nth entry-key 2))
+            board-key (dispatcher/board-data-key org-slug board-slug)
```
bago2k4: @thepug `board-data-key` is not the same as `current-board-key` because the latter distinguish between board and AP based on the router setup.
I think you should pass the board-key as a parameter and get it with the `current-board-key`.
WYT?

src/oc/web/stores/reaction.cljs
```diff
@@ -31,7 +31,10 @@
   [db activity-data reaction reaction-data]
   (let [activity-uuid (:uuid activity-data)
         next-reactions-loading (utils/vec-dissoc (:reactions-loading activity-data) reaction)
-        activity-key (concat (dispatcher/current-board-key) [:fixed-items activity-uuid])]
+        org-slug (utils/post-org-slug activity-data)
+        section-slug (:board-slug activity-data)
+        activity-key (concat (dispatcher/board-data-key org-slug section-slug)
```
bago2k4: @thepug Same goes here

src/oc/web/stores/reaction.cljs
```diff
@@ -50,7 +53,9 @@
         (assoc-in db activity-key updated-activity-data)))))
 
 (defn handle-reaction-to-entry [db activity-data reaction-data]
-  (let [board-key (dispatcher/current-board-key)
+  (let [org-slug (utils/post-org-slug activity-data)
+        section-slug (:board-slug activity-data)
+        board-key (dispatcher/board-data-key org-slug section-slug)
```
bago2k4: @thepug and here

src/oc/web/stores/comment.cljs
thepug: we should probably update both keys (:all-posts , plus the board)

The idea here is that we eventually get rid of this data duplication.

bago2k4: yeah but until we clean up the data duplication we need to at least update the correct data.. for this PR we should stick to the old logic: update the currently visible data path.

bago2k4:
@thepug i was thinking that the solution for comment and reaction stores was to pass the board-key to the action like we do for the section store's actions.
Setting the value for both keys (board and AP) shouldn't cause any problem, at least i think. One of them could be empty since we don't load AP data if you are on a single board page, but setting it with `assoc-in` shouln't raise an exception.
Retesting all the steps to make sure.

thepug:
updated per discussion in slack

PR #485, opened 2018-04-10
Fix activity url

bago2k4:
Bug: when visiting a fullscreen post url i'm redirected to the board and i need to click again on the post (in grid view) to go back to it.

To confirm the bug:
- go to beta
- switch to grid view
- click on a post
- refresh the page
- [x] are you redirected to the board url? Good, well not really

To test the fix:
- locally go to digest
- switch to grid view
- click on a post
- refresh the page
- [x] are you still looking at the fullscreen post page? Good
- go back to digest
- click on the menu of the post
- share url
- paste the url in the broser
- [x] are you redirected to the fullscreen post url and not to the board? Good

### DIFF
--- a/src/oc/web/actions/org.cljs
+++ b/src/oc/web/actions/org.cljs
@@ -57,11 +57,10 @@
            (not (utils/in? (:route @router/path) "secure-activity")))
       ;; Redirect to the first board if at least one is present
       (let [board-to (utils/get-default-board org-data)]
-        (utils/after 10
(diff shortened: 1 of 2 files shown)


thepug:
@bago2k4  This could be unrelated but I get a 404 in incognito mode when I use the share/post url.

thepug:
So , I was logged into another org in incognito.  So the 404 might be the correct behavior.  checking again

PR #487, opened 2018-04-10
Onboard invite teammates and new tooltips [WIP]

bago2k4:
Card: https://trello.com/c/93K5CKQ6

### DIFF
--- a/scss/partials/_onboard_wrapper.scss
+++ b/scss/partials/_onboard_wrapper.scss
@@ -4,7 +4,7 @@ div.onboard-wrapper-container {
     position: absolute;
     top: 0;
     left: 0;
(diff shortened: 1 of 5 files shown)


bago2k4:
@thepug fixed the code issues you found.
About you not seeing the tooltips you need to notice 2 things:
- the add post tooltip is shown only if you have only one post (the default post)
- the invite people tooltip is shown only if there is only one member in the team and show up when the add post tooltip goes away (so if you are an invited user you won't see it)

Ready for review again

thepug:
[NUX: Invite teammates](https://trello.com/c/93K5CKQ6/1249-nux-invite-teammates)

PR #492, opened 2018-04-18
Clean underline text on body paste

bago2k4:
Card: https://trello.com/c/BomxTrBK

To test:
- create a file with HTML extension
- use the content at the bottom of this PR description for the file
- open it in the browser
- copy it
- paste it in our body
- [x] do you not see any underline? Good

Before the last paragraph was keeping the underline formatting since we didn't clean U tags.

```
<html>
  <head>
    <style type="text/css">
      span.und {
        text-decoration: underline;
      }
    </style>
  </head>
  <body>
    <div>
      <p>Hello this is some text with some <span class="und">underlined text</span> in it and <a href="https://google.it">a link</a>.</p>
      <p>Some more text with <span style="text-decoration: underline;">underline set as inline style</span>.</p>
      <p>Some more text withe <u>underline with u tag</u></p>
    </div>
  </body>
</html>
```


bago2k4:
[Clean underline on body paste](https://trello.com/c/BomxTrBK/1283-clean-underline-on-body-paste)

PR #493, opened 2018-04-18
Stream section tag

bago2k4:
Card: https://trello.com/c/4PwIPUUi

To test:
- open AP
- switch to stream if not already
- [x] do you see the section name tag in the right side of the header? Good
- switch to grid view
- [x] do you see the section name tag in the right side of the header? Good
- switch to a single board
- switch to stream
- [x] do you NOT see the section tag? Good
- switch to gred
- [x] do you NOT see the section tag? Good

### DIFF
--- a/scss/mixins/_activity.scss
+++ b/scss/mixins/_activity.scss
@@ -39,6 +39,48 @@ $h2_top_margin: 20;
   }
 }
 
+@mixin new-tag() {
+  div.new-tag {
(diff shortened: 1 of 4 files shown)


bago2k4:
[Section label in header in stream view, too; not just grid view.](https://trello.com/c/4PwIPUUi/1276-section-label-in-header-in-stream-view-too-not-just-grid-view)

PR #496, opened 2018-04-26
Fix user avatar default.

bago2k4:
Card: https://trello.com/c/XXAcNuw6

Test:
- go to signup
- fill in email password
- on profile step add first and last name
- now click on the user avatar and dismiss the FS picker
- [x] is the continue button at the bottom still enabled? Good
- [x] is the avatar still of the color it was initially, not red? Good (if it was red initially no need to check this)
- now upload an avatar to FS
- click again on the new uploaded avatar
- [x] is the continue button at the bottom still enabled? Good
- [x] did the avatar go back to the initial happy face color? Good

### DIFF
--- a/src/oc/web/components/ui/onboard_wrapper.cljs
+++ b/src/oc/web/components/ui/onboard_wrapper.cljs
@@ -154,7 +154,7 @@
                                   (rum/local nil ::temp-user-avatar)
                                   {:will-mount (fn [s]
                                     (user-actions/user-profile-reset)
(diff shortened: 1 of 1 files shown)


bago2k4:
[In NUX, went to add user avatar, then decided not to. Submit button now disabled and can't move on.](https://trello.com/c/XXAcNuw6/1290-in-nux-went-to-add-user-avatar-then-decided-not-to-submit-button-now-disabled-and-cant-move-on)

PR #507, opened 2018-05-03
Fix empty section name

bago2k4:
Card: Can try to create empty section https://sentry.io/opencompany/oc-beta-web/issues/453626704/

Item:
> Can try to create empty section https://sentry.io/opencompany/oc-beta-web/issues/453626704/

To test:
- click + in the navigation
- click Create w/o adding a name
- [x] do you see an error? Good
- type one letter in the section
- [x] did the error go away? Good
- add another char and then 2 or more spaces
- [ ] do you see the error again? Good
- remove all the spaces and add a section name longer then 3 chars
- click Create
- [x] did it work? Good
- click section settings
- delete part of the section name so it's 2 characters long
- click Save
- [x] do you see the same error as before? Good
- change the section name to another longer then 3 chars
- click Save
- [x] did it work? Good

### DIFF
--- a/src/oc/web/components/ui/section_editor.cljs
+++ b/src/oc/web/components/ui/section_editor.cljs
@@ -18,8 +18,6 @@
 
 ;; Private section users search helpers
 
-(def min-section-name-length 3)
-
 (defn filter-user-by-query
   "Given a user and a query string, return true if first-name, last-name or email contains it."
   [user query]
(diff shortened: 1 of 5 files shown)


src/oc/web/components/ui/section_editor.cljs
```diff
@@ -382,16 +391,20 @@
...
-                              section-name (utils/strip-HTML-tags inner-html)
-                              personal-note-node (rum/ref-node s "personal-note")
-                              personal-note (when personal-note-node (.-innerText personal-note-node))
-                              next-section-editing (merge section-editing {:slug utils/default-section-slug
-                                                                           :name section-name})]
-                          (dis/dispatch! [:input [:section-editing] next-section-editing])
-                          (when (fn? on-change)
-                            (on-change next-section-editing personal-note)))}
+            {:on-click (fn [_]
+                        (if (< (count @(::section-name s)) min-section-name-length)
```
thepug: This logic should go into the section action so that it can be reused for other places that create the section.  The action should check the length then dispatch the error.

bago2k4:
@thepug should be good now

PR #517, opened 2018-05-15
Create section copy

bago2k4:
Card: https://trello.com/c/h4qL6IxA

From call discussion: Create as copy of the section editor button is misleading when adding a section from post editing since it doesn't create the section right away but it waits for the post to be posted/saved.
So change the button copy to Done only when showing the section editor from within the section picker (used only when adding/editing a post).

To test:
- go to AP
- click + to add a section
- [x] the copy of the button at the bottom is Create? Good
- nav to a single section
- click the section settings button
- [x] the copy of the button at the bottom is Save? Good
- click Compose
- click on the section name
- click Add a new section
- [x] the copy of the button at the bottom is Done? Good

### DIFF
--- a/src/oc/web/components/ui/section_editor.cljs
+++ b/src/oc/web/components/ui/section_editor.cljs
@@ -110,7 +110,7 @@
                                 (events/unlistenByKey @(::click-listener s))
                                 (reset! (::click-listener s) nil))
                               s)}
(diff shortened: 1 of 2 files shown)


bago2k4:
[Create section button copy change](https://trello.com/c/h4qL6IxA/1317-create-section-button-copy-change)

PR #528, opened 2018-05-28
Notifications restyle

bago2k4:
Card: https://trello.com/c/wnvEtf7r

Abstract: https://share.goabstract.com/ad3dbe9b-2636-4565-9562-18a9fdb186df

This implements a notification system. It mounts a new React root at the top level to make sure the notifications will always be visible, no matter what the app React root is.
Limitations: for now implements only the server error and app version error notifications but it's ready to show any kind of notifications. For those who needs an UNDO or CANCEL action that's still to be worked but it should be pretty quick.
Notifications can have an expire time, default to 5 seconds, 0 means it never goes away.
Also this avoids showing the same notifications multiple times by using an id to remove duplicates, if 2 notifications are shown with the same id the one with bigger expire time is kept, if one has 0 that's kept (because it's like infinite).

To test:
- start the app
- stop storage
- [x] do you see a server error notification? Good
- [x] can you make it disappear by clicking the X? Good
- [x] do you see only one even if the console show 2 times the `:show-notification` action? Good
- now open a staging window
- deploy this to staging (no matter if you are deploying the same code, we need staging to update the code
- when deploy is finished you still need to wait for the local app auth-settings expire time (it's 20 mins at most)
- now nav to another section or AP
- [x] do you see the app update banner appearing? Good


bago2k4:
[Notifications restyle](https://trello.com/c/wnvEtf7r/1336-notifications-restyle)

PR #532, opened 2018-06-04
Disable fullstory in local development and on staging.

bago2k4:
### Review with https://github.com/belucid/open-company-infrastructure/pull/67

To test:
- run local dev
- [x] check that `init_fullstory` is not called
- [x] check that no call is done to fullstory
- deploy to staging
- [x] check that `init_fullstory` is not called
- [x] check that no call is done to fullstory
- merge
- deploy to beta
- [ ] check that `init_fullstory` IS called
- [ ] check that we are sending fullstory the bundles


site/oc/pages.clj
```diff
@@ -873,8 +873,6 @@
           [:script {:type "text/javascript" :src "/lib/autotrack/autotrack.js"}]
           [:script {:type "text/javascript" :src "/lib/autotrack/google-analytics.js"}]
           (google-analytics-init)
-          [:script {:type "text/javascript" :src "/lib/fullstory.js"}]
```
thepug: The script tag seems missing below

bago2k4: @thepug do you see an error? You shouldn't, i removed fullstory.js only from the local dev app-shell, that was using it directly. In `prod-app-shell` we use the version inside oc-assets.js that is compiled using `compile_assets.sh`.

PR #533, opened 2018-06-04
Fix sticky post body

bago2k4:
BUG: sometimes when clicking Compose button you see that it's populated with the body of the last post you wrote or edited.

Problem: the bug is caused by a race condition btw the view dismiss and the autosave interval we used to avoid losing the user entered text.

Reproduce: it's very hard to reproduce, i was able to do it by checking the indexedDB from the devtools. Since we save the body every 5 seconds, follow these steps:
- click Compose
- type something in the headline
- type something in the body
- looking at indexedDB in dev tools you will be able to see when the autosave take place
- when autosave happens count 4 seconds
- hit Post
with this steps i was able to get the sticky cache every now and then.

Solution: repeat the steps above with the current branch. The problem is that we are removing the interval used to autosave when the entry-edit view is dismissed, meaning that there is a fraction on time btw the save (that removes the last cache) and the view dismiss where, if the autosave happens to fall in this time, you will find the post body when you get back to composing.

### DIFF
--- a/src/oc/web/components/entry_edit.cljs
+++ b/src/oc/web/components/entry_edit.cljs
@@ -37,6 +37,11 @@
 
 ;; Local cache for outstanding edits
 
+(defn remove-autosave [s]
+  (when @(::autosave-timer s)
+    (.clearInterval js/window @(::autosave-timer s))
+    (reset! (::autosave-timer s) nil)))
+
 (defn autosave [s]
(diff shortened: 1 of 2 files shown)


bago2k4:
[Old post returns as a "draft" when trying to compose something new](https://trello.com/c/wJIMiC5g/1345-old-post-returns-as-a-draft-when-trying-to-compose-something-new)

PR #535, opened 2018-06-07
Preload sections fix

bago2k4:
Card: https://trello.com/c/ha8ZFQCt

Bug: create a draft, switch to draft board, refresh. Notice that the compose button is disabled. The problem is that the drafts board is readonly (you can't create post directly inside it). The problem is that we don't preload the sections when you load the app from a section instead of AP.

This should fix the above scenario by making sure the sections are all loaded when the org data are loaded.

To test:
- go to the drafts board
- refresh the app
- [x] can you click the compose button? Good
- click it
- click on the board name
- [x] do you see all the sections you can add a post to? Good
- repeat the above from a normal section (not drafts)
- repeat the above from AP

### DIFF
--- a/src/oc/web/actions/org.cljs
+++ b/src/oc/web/actions/org.cljs
@@ -68,6 +68,7 @@
   ;; Check the loaded org
   (let [ap-initial-at (:ap-initial-at @dis/app-state)
         boards (:boards org-data)]
+    (sa/load-other-sections (:boards org-data))
     (cond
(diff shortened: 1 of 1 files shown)


bago2k4:
[Load app in Drafts, and compose sees it as a viewer only](https://trello.com/c/ha8ZFQCt/1347-load-app-in-drafts-and-compose-sees-it-as-a-viewer-only)

PR #551, opened 2018-06-12
Client Change services defect

bago2k4:
Card: https://trello.com/c/NCRYPnK1/1371-client-change-services-defect

Follow the steps in the card to reproduce the issue and check if it's fixed.


bago2k4:
[Client Change services defect](https://trello.com/c/NCRYPnK1/1371-client-change-services-defect)

PR #553, opened 2018-06-13
Fix navigation height calc

bago2k4:
Fix the navigation button, make sure the invite people button is in the right place when the screen is not tall enough to have the absolutely positioned navigation sidebar and when it's not tall enough.

To test:
- use an org with some sections
- resize the screen height
- make sure the invite people button is always in the right place and it's always reachable (scrolling or not)


bago2k4:
[Draft shenanigans - new section and delete section draft bugs](https://trello.com/c/JvEN5yD0/1363-draft-shenanigans-new-section-and-delete-section-draft-bugs)

PR #559, opened 2018-06-18
Fix post removed

bago2k4:
Card: https://trello.com/c/esSCYxRj

Item:
> Update 404 copy: 

**Post Not Available**
_Looks like you don't have access to the post or it doesn't exist._

To test:
- visit /carrot/dsa/post/1234-1234-1234 when logged in
- [x] do you see the copy above? Good
- visit the same url when not logged in
- [x] do you see the login prompt/wall instead? Good

### DIFF
--- a/src/oc/web/actions/org.cljs
+++ b/src/oc/web/actions/org.cljs
@@ -93,12 +93,7 @@
         ; The board wasn't found, showing a 404 page
         (if (= (router/current-board-slug) utils/default-drafts-board-slug)
           (utils/after 100 #(sa/section-get-finish utils/default-drafts-board))
-          (if (router/current-activity-id) ;; user is asking for a specific post
-            (if (jwt/jwt)
-              ;; Show the activity removed
-              (dis/dispatch! [:input [:show-activity-removed] true])
-              ;; Show the activity not found/login screen
-              (dis/dispatch! [:input [:show-activity-not-found] true]))
+          (when-not (router/current-activity-id) ;; user is not asking for a specific post
             (router/redirect-404!))))
       ;; Board redirect handles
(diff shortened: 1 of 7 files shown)


belucid:
@bago2k4 

This sequence doesn't work:

- Logged in user
- Go to 404 URL specified in testing steps
- Click the go to homepage button
- After All Posts loads, click the browser back button

Rather than the 404 again, you get infinite shaky carrot

bago2k4:
@belucid i think i need more details on this. If i click back after i clicked on Go to homepage i go back to the post removed screen.
That looks good to me. I tried removing the cookie for the last visited org slug that is responsible for redirecting you directly to /last-org/all-posts or instead to /login if the cookie is not found. In both cases if i click back i don't get stuck even though i get different behaviour (when i'm redirected to /login first then i go to the org dashboard, if i click back here i go to login again that redirects me again to the dashboard).

bago2k4:
@belucid try again, i changed the logic how we show those walls.
Before i was setting the needed to show the wall in the app-state. Since we don't reload all the app data when an internal route happens the flag was not set again when going back.
Now the check happens in the org-dashboard component. I don't like it very much because it keeps the logic in the component but it's the only thing i could with this state of the art.

thepug:
@bago2k4 conflickts on this branch

bago2k4:
@thepug conflict resolved

thepug:
> do you see the login prompt/wall instead?
@bago2k4 is this the correct behavior? 
http://g.recordit.co/bP7T3AdyIE.gif

bago2k4:
@thepug good catch, locally it wasn't working because i was not passing the jwt as a derivative to the component but there was the namespace with the same name that was checking true.
On staging it worked because the names are compiled.
Good catch anyway.

PR #573, opened 2018-07-06
Refactor post data in app state.

thepug:
This changes almost all things related to posts, including comments and reactions.

I think the best way to test this is look at the code change and manually test what behavior the code change affects.

You do want to make sure that the sections on the left display the expected results.   

On each clicked section you want to be able to create a new entry.


thepug:
[Single Post state de-dupe](https://trello.com/c/zQbkrkTF/1384-single-post-state-de-dupe)

bago2k4:
@thepug an early feedback FYI. Mobile seems to have lost the list of posts, at least if you open the app to AP. If you switch to a board and then back to AP everything works.
Also i see 3 warnings referring to wrong number of arguments for `activity-key` fn when dispatching `:activity-move` (in oc.web.stores.activity).
No rush, just a reminder

bago2k4:
Note mostly for me, the `oc.web.actions.nux/check-nux` doesn't work anymore because `oc.web.dispatcher/board-data` is empty when on AP or MS. Before it wasn't.

thepug:
@bago2k4 I have addressed the storing of the container information.  For pre-caching data I don't think we think about that now.  For one, since we load every board we have all the posts anyway.  I think we want to deal with the server side API calls later.  Probably when we add paging to the boards.

thepug:
In order to address some of the loading issues  I chose to skip loading entries on the initial section load. We only seem to need the section information (name, slug, links, etc)  
https://github.com/open-company/open-company-web/pull/573/commits/02f7e337238ed5c06e768a68016615d6fb1fd950

src/oc/web/stores/reaction.cljs
```diff
@@ -156,9 +155,8 @@
   [db interaction-data add-event?]
   (let [item-uuid (:resource-uuid interaction-data)
         comments-key (get @reactions-atom (make-comment-index item-uuid))
-        all-posts-key (assoc comments-key 2 :all-posts)
-        ;; look for data in the board and in AP key
-        keys (remove nil? [comments-key all-posts-key])
+        ;; look for data in the board
+        keys (remove nil? [comments-key])
```
bago2k4: remove nil? is not needed anymore since coments-key is the only key used and will always be not null

PR #582, opened 2018-07-17
Fix WST requests

bago2k4:
BUG: if you check the CS logs starting the web app on mainline right now you get this error every now and then:
```
18-07-17 15:51:25 bagoPro.local INFO [oc.change.async.persistence:?] - Who read request for: () by: c891c6cf-a683-477c-b8a6-69a611fe37eb
18-07-17 15:51:25 bagoPro.local ERROR [oc.change.async.persistence:204] -
                                                               java.lang.Thread.run              Thread.java:  745
                                 java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  617
                                  java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1142
                                                                                ...
                                                  clojure.core.async/thread-call/fn                async.clj:  441
oc.change.async.persistence/persistence-loop/fn/state-machine--auto--/fn/inst-29242          persistence.clj:  202
                                                                                ...
                             oc.change.async.persistence/handle-persistence-message          persistence.clj:   85 (repeats 2 times)
                          oc.change.async.persistence/handle-persistence-message/fn          persistence.clj:   85
                                        oc.change.resources.read/eval29087/retrieve                 read.clj:   37
clojure.lang.ExceptionInfo: Input to retrieve does not match schema:

                            	   [(named (not (oc.lib.schema/unique-id? ())) item-id)]


     error: [(named (not (oc.lib.schema/unique-id? ())) item-id)]
    schema: [{:schema (pred oc.lib.schema/unique-id?),
              :optional? false,
              :name item-id}]
      type: :schema.core/error
     value: [()]
```
As you can see the problem is that the client sends requests with an empty list of items. This happens because to avoid loading multiple times the same data we filter the list of item ids with the list of counts we already have. After the initial loads this difference is often empty since we have already loaded all the count data.

To test:
- start this branch
- go to AP
- [x] check that you get the reads count
- hover on some posts views count to load the reads list
- [x] check that you get the list of reads
- switch to a section
- [x] check that you still have the reads data
- hover on some posts views count to load the reads list
- [x] check that you get the list of reads
- refresh on the section
- [x] check that you see the reads count for all the posts
- hover on some posts
- [x] check that you see the reads list
- switch to AP (no refresh)
- [x] check that you see the reads count for the posts that are not in the section you were before
- hover on some posts from the section you came from and not
- [x] check that you see the reads list for all of them

### DIFF
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -36,11 +36,17 @@
 
 ;; Reads data
 
-(defn request-reads-data [item-id]
+(defn request-reads-data
+  "Request the list of readers of the given item."
+  [item-id]
   (api/request-reads-data item-id))
 
-(defn request-reads-count [item-ids]
-  (api/request-reads-data item-ids))
+(defn request-reads-count
+  "Request the reads count data only for the items we don't have already."
+  [item-ids]
+  (let [cleaned-ids (au/clean-who-reads-count-ids item-ids (dis/activities-read-data))]
+    (when (seq cleaned-ids)
+      (api/request-reads-data cleaned-ids))))
 
 ;; All Posts
 (defn all-posts-get-finish [from {:keys [body success]}]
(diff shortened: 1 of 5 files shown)


belucid:
@bago2k4 this all checks out except in the case of an AP refresh on FF where we don't get any initial read requests. This seems to be quite sporadic and related to the WS race condition.

PR #590, opened 2018-08-06
Fix onboard overlay

bago2k4:
Stuart's bug: 
> chrome/desktop. created a newco with email.
> first thing I did was create a new section. I didn’t dismiss the yellow tip. just went straight to create a new section.
> my email was stuartlevinson+myeyedr@gmail.com, so it was a new email

To test:
- replicate the scenario
- [ ] do you NOT see the onboard overlay after adding a new section? Good


belucid:
@bago2k4 

repeat Stuart's steps
then hit refresh

http://take.ms/FZhUt

PR #594, opened 2018-08-20
Bugfix: Slack team info lost

bago2k4:
Bug: i cleaned my DBs and signed up with Slack from scratch. I found out that the team name and logo where not inherited anymore but they were still set in the team info.

To test:
- clean your DBs
- signup with Slack
- [x] do you see the slack team name and logo in the team lander page? Good
- add an email domain
- [x] do you see the red borders only when there is a not valid domain in the field? Good
- [x] can you submit with an email domain and is that kept in settings? Good
- sign out
- signup with google (using an account that inherit the domain)
- [x] do you see the email domain set in the proper field in the team lander page? Good
- enter a team name
- add a logo
- proceed
- [x] check that all the info above are kept

@thepug check how :update differ from :input in `dispatch!`, you need to pass a function to :update not a single value: https://github.com/open-company/open-company-web/blob/mainline/src/oc/web/actions.cljs#L24
Telling you this because of a change you did in Email domain NUX PR and i didn't catch.

### DIFF
--- a/src/oc/web/components/ui/onboard_wrapper.cljs
+++ b/src/oc/web/components/ui/onboard_wrapper.cljs
@@ -253,17 +253,23 @@
   ;; Load the list of teams if it's not already
   (team-actions/teams-get-if-needed)
   (let [org-editing @(drv/get-ref s :org-editing)
-        teams-data @(drv/get-ref s :teams-data)]
-    (when (and (zero? (count (:name org-editing)))
-               (zero? (count (:logo-url org-editing)))
-               (seq teams-data))
+        teams-data @(drv/get-ref s :teams-data)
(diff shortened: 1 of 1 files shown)


src/oc/web/components/ui/onboard_wrapper.cljs
```diff
@@ -395,26 +395,12 @@
                  :type "text"
                  :auto-capitalize "none"
                  :pattern "@?[a-z0-9.-]+\\.[a-z]{2,4}$"
+                 :value (:email-domain org-editing)
```
thepug: Using `:value` causes a bug which is why I switched to `(set~ (.-value field)` on `:did-mount`

http://g.recordit.co/IOJ257X59t.gif

PR #596, opened 2018-08-23
Auto draft

thepug:
Requirements doc: https://docs.google.com/document/d/1TTJdi_ajlrUkotDhVL1qT1smuSFH22zzcAoqdBkPvjo/edit
Trello: 
https://trello.com/c/Sc2g1qL4/1447-saving-saved-server-side-not-just-with-cookie

When a user creates a draft, these changes will automatically save any changes made. Both on the server side and in the client.

Depends on https://github.com/open-company/open-company-storage/pull/142

Test steps:

Test edit view
- click 'New'
- type in some changes
- [ ] Do you see the 'Saving..' 'Saved' message?
- close the browser window
- [ ] after reloading you should see a draft with your changes
***

- edit the draft
- add some changes
- [ ] Do you see the 'Saving..' 'Saved' message?
- click the x
- click 'lose changes'
- [ ] is the original post restored?

***
- click 'New'
- add some changes
- [ ] Do you see the 'Saving..' 'Saved' message?
- click x
- [ ] is the post deleted?
***

- click 'New'
- make some changes
- click save
- [ ] is there a draft?

***
- click 'New'
- make some changes
- click Post
- [ ] is there a post?

***
- click 'New'
- save a draft
- edit the draft
- click Post
- [ ] is there a post?

***
Test full screen edit
Normal full screen edit should function as before. Just go through some edits to ensure it still works.


src/oc/web/actions/activity.cljs
```diff
@@ -239,13 +242,30 @@
...
         cache-key (get-entry-cache-key (:uuid activity-data))]
     (uc/set-item cache-key entry-map
      (fn [err]
        (when-not err
+         (when (and (not= "published" (:status activity-data))
+                    (:has-changes activity-data)
+                    (not (:auto-saving activity-data)))
+           ;; dispatch that you are auto saving
+           (dis/dispatch! [:update [:entry-editing]
+                           #(merge % activity-data {:auto-saving true})])
```
bago2k4: You need to use `edit-key` here instead of `:entry-editing` or you'll always do it only for the `entry-edit` component.

Also much simpler to use:
`(dis/dispatch! [:input [:entry-editing :auto-saving] true])`
sorry to bother on these :input and :update but i am trying to make more clear how they works since they are simple but confusing.

src/oc/web/actions/activity.cljs
```diff
@@ -239,13 +242,30 @@
...
+           (entry-save activity-data section-editing
+             (fn [entry-data-saved edit-key-saved {:keys [success body status]}]
+               (when success
+                 (remove-cached-item (:uuid activity-data))
+                 (let [entry-saved (assoc (json->cljs body) :auto-saving false)]
+                   ;; merge with entry editing and only save once we have a uuid
+                   (dis/dispatch! [:update [:entry-editing]
```
bago2k4: You need to use `edit-key-saved` here instead of `:entry-editing`.

src/oc/web/actions/activity.cljs
```diff
@@ -321,40 +341,50 @@
...
-                ;; Board name exists
-                (board-name-exists-error fixed-edit-key)
-                (create-update-entry-cb fixed-edited-data fixed-edit-key response)))))
-        (let [org-slug (router/current-org-slug)
-              entry-board-data (dis/board-data @dis/app-state org-slug (:board-slug fixed-edited-data))
-              entry-create-link (utils/link-for (:links entry-board-data) "create")]
-          (api/create-entry fixed-edited-data fixed-edit-key entry-create-link create-update-entry-cb))))
-    (dis/dispatch! [:entry-save fixed-edit-key])))
+(defn entry-save
+  ([edited-data section-editing]
```
bago2k4: This function needs an `edit-key` as parameter so we know what key to use.

For example the component entry-edit uses `:entry-editing` to save the editing data but the new cmail uses `:cmail-data`. We use different keys because this app-state key is also used to decide which component we need to show.
In `org-dashboard` we show `entry-edit` if `:entry-editing` is not nil and `cmail` if `:cmail-data` is not nil.

src/oc/web/actions/activity.cljs
```diff
@@ -321,40 +341,50 @@
...
+     (entry-save edited-data section-editing create-update-entry-cb))
```
bago2k4: And you need to pass the `edit-key` parameter along.

src/oc/web/actions/activity.cljs
```diff
@@ -321,40 +341,50 @@
...
+
+  ([edited-data section-editing entry-save-cb]
```
bago2k4: Same as above you need to accept an `edit-key` parameter.
You can also add a signature w/o it that defaults to `:entry-editing` if you want.

src/oc/web/actions/activity.cljs
```diff
@@ -321,40 +341,50 @@
...
+     (let [fixed-edited-data (assoc edited-data :status (or (:status edited-data) "draft"))
+           fixed-edit-key :entry-editing]
```
bago2k4: And here you need:
`fixed-edit-key (or edit-key :entry-editing)`

src/oc/web/components/entry_edit.cljs
```diff
@@ -516,6 +522,10 @@
                          :position "top"
                          :default-field-selector "div.entry-edit-modal div.rich-body-editor"
                          :container-selector "div.entry-edit-modal"})
+          (when (= true (:auto-saving entry-editing))
```
bago2k4: Do we really need to check if it's equal to `true`? If so you can use `true?` that is a big more clean.
https://clojuredocs.org/clojure.core/true_q

src/oc/web/components/entry_edit.cljs
```diff
@@ -516,6 +522,10 @@
...
+            [:div.saving-saved "Saving..."])
+          (when (= false (:auto-saving entry-editing))
```
bago2k4: Same here, you can use `false?` https://clojuredocs.org/clojure.core/false_q

src/oc/web/actions/activity.cljs
thepug: So I used the merge function because I need the links information and the revision information if the entry is auto saved.  Using the suggested input dispatch doesn't add that information.

src/oc/web/actions/activity.cljs
```diff
@@ -239,13 +242,30 @@
...
     (uc/set-item cache-key entry-map
      (fn [err]
        (when-not err
+         (when (and (not= "published" (:status activity-data))
+                    (:has-changes activity-data)
+                    (not (:auto-saving activity-data)))
+           ;; dispatch that you are auto saving
+           (dis/dispatch! [:update [edit-key]
+                           #(merge % activity-data {:auto-saving true})])
+           (entry-save edit-key activity-data section-editing
```
bago2k4: Saving the entry with the server on each edit can be a problem because on the first edit we don't have a uuid and after the first save we do. This will compromise the local save to cache that is based on the entry uuid.
In the entry-save callback we need to check if the entry had a uuid, if not remove the cached item and save a new one with the proper uuid.

src/oc/web/actions/activity.cljs
```diff
@@ -239,13 +242,30 @@
...
+             (fn [entry-data-saved edit-key-saved {:keys [success body status]}]
+               (when success
+                 (remove-cached-item (:uuid activity-data))
```
bago2k4: If you remove the cached item every time you save the local cookie won't be saving anymore. You need to remove the cache only in the case i described above.

src/oc/web/actions/activity.cljs
bago2k4: Also check that the revision-id is properly saved after you start editing a post without a uuid and it's saved the first time, as i mentioned in Slack i started a new post, edited it 3 times, closed out of the edit and the revert was for version 1 instead of 0 (probably because after the first save it replaced the starting revision).

src/oc/web/actions/activity.cljs
```diff
@@ -239,14 +243,37 @@
       (dis/dispatch! [:modal-editing-activate]))
     (dis/dispatch! [:modal-editing-deactivate])))
 
+
+(declare entry-save)
+
 (defn entry-save-on-exit
```
bago2k4: A `docstring` on this function and also some inline comments when you set/check/remove the initial-revision atom and/or the local cache would be great @thepug

src/oc/web/actions/activity.cljs
```diff
@@ -239,14 +243,37 @@
...
+           ;; dispatch that you are auto saving
+           (dis/dispatch! [:update [edit-key]
+                           #(merge % entry-map {:auto-saving true})])
+           (entry-save edit-key entry-map section-editing
+             (fn [entry-data-saved edit-key-saved {:keys [success body status]}]
+               (when success
+                 (let [entry-saved (assoc (json->cljs body) :auto-saving false)]
+                   (when (and (nil? (:uuid entry-map))
+                              (:uuid entry-saved))
+                     (remove-cached-item (:uuid entry-map)))
```
bago2k4: For example here, why do we remove the cached item?

src/oc/web/actions/activity.cljs
```diff
@@ -239,14 +243,37 @@
...
+                   (when (nil? (get @initial-revision (:uuid entry-saved)))
+                     (reset! initial-revision
```
bago2k4: And why we reset the initial-revision here only if...

src/oc/web/actions/activity.cljs
```diff
@@ -261,8 +288,9 @@
       (router/nav! (oc-urls/entry org-slug board-slug (:uuid activity-data))))
     (save-last-used-section board-slug)
     (refresh-org-data)
-    ; Remove saved cached item
+    ;; Remove saved cached item
     (remove-cached-item initial-uuid)
+    (reset! initial-revision (dissoc @initial-revision (:uuid activity-data)))
```
bago2k4: Here too..

src/oc/web/actions/activity.cljs
```diff
@@ -279,6 +307,7 @@
         saved-activity-data (first (vals (:fixed-items fixed-board-data)))]
     (save-last-used-section (:slug fixed-board-data))
     (remove-cached-item (:uuid activity-data))
+    (reset! initial-revision (dissoc @initial-revision (:uuid activity-data)))
```
bago2k4: And here..

src/oc/web/actions/activity.cljs
```diff
@@ -321,47 +350,59 @@
     (fn [{:keys [status success body]}]
       (dis/dispatch! [:activity-get/finish status (router/current-org-slug) (json->cljs body) nil]))))
 
-(defn entry-clear-local-cache [item-uuid edit-key]
+(declare entry-revert)
+
+(defn entry-clear-local-cache [item-uuid edit-key item]
   (remove-cached-item item-uuid)
+  ;; revert draft to old version
```
bago2k4: This is good, i would add why we revert only in some cases?

src/oc/web/actions/activity.cljs
```diff
@@ -321,47 +350,59 @@
...
+               (fn [{:keys [success status body] :as response}]
+                 (if (= status 409)
+                   ;; Board name exists
+                   (board-name-exists-error fixed-edit-key)
+                   (entry-save-cb fixed-edited-data fixed-edit-key response)))))
+           (let [org-slug (router/current-org-slug)
+                 entry-board-data (dis/board-data @dis/app-state org-slug (:board-slug fixed-edited-data))
+                 entry-create-link (utils/link-for (:links entry-board-data) "create")]
+             (api/create-entry fixed-edited-data fixed-edit-key entry-create-link entry-save-cb))))
+       (dis/dispatch! [:entry-save fixed-edit-key]))))
```
bago2k4: This all would be good to have some comments at least to understand what's doing in the various cases (create new entry, create new entry with new board, update existing entry etc..).
I know it's not your code, you can avoid it if you don't have enough time..

src/oc/web/actions/activity.cljs
```diff
@@ -455,6 +497,19 @@
   (api/share-activity activity-data share-data (or share-cb activity-share-cb))
   (dis/dispatch! [:activity-share share-data]))
 
+(defn entry-revert [revision-id entry-editing]
```
bago2k4: Docstring and inline comments here would be great

thepug:
[Saving, saved server-side, not just with cookie](https://trello.com/c/Sc2g1qL4/1447-saving-saved-server-side-not-just-with-cookie)

belucid:
@thepug 

Following the initial test steps above:

```
click the x
click 'lose changes'
```

I never get a prompt when I click the X. It just closes and the changes are gone.

thepug:
@belucid I thought this was the expected behavior for this branch. I can add the lose changes dialog on all x clicks.  This will all change in cmail, right?  We can also wait and get cmail out for this.

> @thepug

> Following the initial test steps above:

> click the x
> click 'lose changes'
> I never get a prompt when I click the X. It just closes and the changes are gone.

bago2k4:
Closing since this was merged here https://github.com/open-company/open-company-web/pull/611

PR #608, opened 2018-09-06
Last section standing

bago2k4:
Card: https://trello.com/c/QNkwf9i1

Abstract: https://projects.invisionapp.com/share/WNNVHKIT7XH

To test:
- start with at least 2 sections
- delete a section
- [x] do you see the section disappear from left nav and you are redirected to AP? Good
- click New
- [x] do you NOT see the deleted section in the sections list picker? Good
- now try to delete the last section
- [x] could you NOT delete it? Good
- [x] did you see the tooltip saying that you can't delete the last remaining section? Good


bago2k4:
[Last section standing](https://trello.com/c/QNkwf9i1/1466-last-section-standing)

src/oc/web/components/ui/section_editor.cljs
```diff
@@ -371,25 +373,33 @@
...
+                                                 (alert-modal/hide-alert)
+                                                 (dismiss))})))
+               :data-toggle "tooltip"
+               :data-placement "top"
+               :data-container "body"
+               :title (if last-section-standing
+                       "You cannot delete the last remaining section."
+                       "Delete this section and all its posts.")
+               :class (when last-section-standing "disabled")}
               "Delete section"])
```
thepug: Should we remove the delete button completely?

Also, I can delete all the sections.
http://recordit.co/OIfp05eEN5

bago2k4: We don't want to remove the delete button because we want to let the users know why they can't delete the section.. Fixing now the issue

bago2k4:
@thepug so the issue you found where you could delete the last section is fixed.
The point was that one of your sections wasn't really deleted before because you or another user had drafts in it.
When you try to do that we turn on the draft flag for that section, it means that it's a write only section at the moment, it's not visible in the sections list but it's visible in the sections for editing.
This is to avoid deleting drafts of other users.
To fix the issue i excluded the sections marked as draft from the last section standing count, meaning: you can't delete the last visible section, not the last overall section.

PR #635, opened 2018-09-18
Save when closing and auto save after adding an attachment.

thepug:
Test steps:

- open up cmail
- add an attachment
- [x] does it auto save?  (Saving... , Saved)
- add a header and try to click x before the draft is saved
- [x] does it still save the data?


thepug:
[Make sure to auto save when image/attachment is added.  Make sure we auto save when clicking the x](https://trello.com/c/G9mlJxdD/1475-make-sure-to-auto-save-when-image-attachment-is-added-make-sure-we-auto-save-when-clicking-the-x)

src/oc/web/actions/activity.cljs
```diff
@@ -338,10 +338,12 @@
   (dis/dispatch! [:entry-modal-save]))
 
 (defn add-attachment [dispatch-input-key attachment-data]
-  (dis/dispatch! [:activity-add-attachment dispatch-input-key attachment-data]))
+  (dis/dispatch! [:activity-add-attachment dispatch-input-key attachment-data])
+  (dis/dispatch! [:input [:cmail-data :has-changes] true]))
```
bago2k4: @thepug you need to use the passed `dispatch-input-key` or this will work only for Cmail and not on mobile when using the entry-edit component.

src/oc/web/actions/activity.cljs
```diff
@@ -338,10 +338,12 @@
...
 
 (defn remove-attachment [dispatch-input-key attachment-data]
-  (dis/dispatch! [:activity-remove-attachment dispatch-input-key attachment-data]))
+  (dis/dispatch! [:activity-remove-attachment dispatch-input-key attachment-data])
+  (dis/dispatch! [:input [:cmail-data :has-changes] true]))
```
bago2k4: Same here @thepug

src/oc/web/components/rich_body_editor.cljs
```diff
@@ -86,6 +86,7 @@
   (let [editable (or editable (get-media-picker-extension s))]
     (.saveSelection editable))
   (dis/dispatch! [:input [:media-input :media-video] true])
+  (dis/dispatch! [:input [:cmail-data :has-changes] true])
```
bago2k4: Why you set cmail has-changes here? This lines is showing the overlay to collect the video url (`media-video-modal` component).

The add video and add images inside the ME actions always end up changing the body of the entry, that HTML node has a listener that set the `has-change` when something changes inside it.

I don't think we need to make sure we autosave here, maybe when `media-video-add` and `media-photo-add-if-finished` are called but as i said the body automatically set has-changes to true when something changes inside.

Why did you add this here?

thepug:
@bago2k4 addressed your feedback https://github.com/open-company/open-company-web/pull/635/commits/332916e7780cbd48bf9e6bdcc6ac2f2f15d91090

PR #637, opened 2018-09-19
New section cmail

thepug:
Needs https://github.com/open-company/open-company-storage/pull/148

Test steps:

- click new
- before adding any other data change the section
- click add new section and add a new one
- [x] after adding the new section the post should auto save
- [x] can you add a header or body and it correctly save?

- click new
- use an existing section
- add a header
- [x] does the post auto save correctly?

### DIFF
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -253,12 +253,13 @@
              (fn [entry-data-saved edit-key-saved {:keys [success body status]}]
                (when success
                  (let [json-body (json->cljs body)
-                       board-data (if (:slug json-body)
+                       board-data (if (:entries json-body)
                                     (au/fix-board json-body)
                                     false)
-                       entry-pre-merge (if (:fixed-items json-body)
+                       fixed-items (:fixed-items board-data)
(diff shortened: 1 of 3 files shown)


thepug:
[Autosave fails when creating a new section.](https://trello.com/c/sJz4u8KV/1478-autosave-fails-when-creating-a-new-section)

bago2k4:
@thepug  you PRs checks out, i thought about some potential race condition though and i pushed 3 commits

one is very simple and is needed to show the Saving… label also when we are saving and not only when we are waiting to save (not only when has-changes is true but also when auto-saving is true)

another is to avoid saving the auto-saving flag in the local cookie, if we do save it and we pick up the entry from there autosave won’t never run because that is always true (we turn that off at the end of autosave)

last one is to avoid overriding user edits with the autosave return value, right now i do this:
when autosave starts i remove has-changes and add auto-saving flags to avoid another autosave for the same data, when autosave finishes if it failed i put back auto-saving false and has-changes true, this way autosave will be retried next time. If autosave succeeded i simply remove auto-saving and don’t touch anything else (to avoid overriding user edits)

waiting for your check to merge this

PR #638, opened 2018-09-19
Fix empty avatar

bago2k4:
From:

Items:
> For users that don’t upload a workplace icon, please use this design so that they’re prompted to add one
> Both in the settings drop down and settings modal

To test:
- use a user w/o logo in an org w/o logo
- go to user profile
- [ ] do you see the green circle instead of the jelly face? Good
- upload a profile logo
- [ ] do you see the profile logo? Good
- go to org settings
- [ ] do you see the green square with a plus in the middle? Good
- upload a logo
- [ ] do you see the logo? Good

### DIFF
--- a/scss/partials/_org_settings.scss
+++ b/scss/partials/_org_settings.scss
@@ -56,37 +56,31 @@ div.org-settings {
 
       div.org-settings-header-avatar {
         float: left;
-        border: 1px solid #D8DBDF;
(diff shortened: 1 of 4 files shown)


bago2k4:
[Fix empty avatars](https://trello.com/c/oc6HSJPw/1485-fix-empty-avatars)

PR #646, opened 2018-09-25
Use search result limit since we don't have paging

thepug:
Test steps

- Search for a term resulting in more than 20 results. An author's name is best for this.
- [x] does the result count only ever show 20?
- [x] after clicking show more does it go away and display 20 results?


thepug:
[Show more in search gathers all results but it isn't clear that there are more available.](https://trello.com/c/wYbVVC1q/1477-show-more-in-search-gathers-all-results-but-it-isnt-clear-that-there-are-more-available)

src/oc/web/components/search.cljs
```diff
@@ -70,14 +73,17 @@
                                   s)}
   [s]
   (let [search-results (drv/react s store/search-key)
-        search-active? (drv/react s store/search-active?)]
+        search-active? (drv/react s store/search-active?)
+        result-count (if (< store/search-limit (:count search-results))
+                       store/search-limit
+                       (:count search-results))]
```
bago2k4: @thepug Why do you repeat this result-count calculation? can you pass it to the `results-header` component to avoid recalc it there too?

thepug: that would just make sense :)

thepug:
@bago2k4 addressed your feedback

PR #648, opened 2018-09-25
Fix for missing search result.

thepug:
Only after finishing the NUX and searching for a post will this occur.  The issue is that the app state doesn't have any org information in the `:orgs` key.  This would cause the org dashboard component to say the post was missing because it thought the org was missing.

To test -

- finish the nux
- search for sample data
- click on the result
- [ ] without this change it will say the post is missing
- [ ] with this change the post will be displayed

- test that all other org related things are functioning


thepug:
[Search [sample] posts doesn't work](https://trello.com/c/ErOVxOTy/1489-search-sample-posts-doesnt-work)

thepug:
need to load entry point after NUX

PR #656, opened 2018-10-02
Don't add drafts to must see data in app state.

thepug:
This change fixes the issue where a draft will show up when viewing the must see section.

To test:

- navigate to must see
- click new
- type a headline, body , and make sure it is toggled as must see
- [x] Does the draft show up in the list of must see posts?  If not , all is good. You can revert the change to test that it does show up.


thepug:
[Drafts marked "must see" should not appear in Must see until posted](https://trello.com/c/ijjCfdC0/1492-drafts-marked-must-see-should-not-appear-in-must-see-until-posted)

PR #660, opened 2018-10-04
Support links

belucid:
This branch adds support links that invoke Drift in the dropdown menu and in the left nav.

To test:

- Temporarily update the `DRIFT_KEY` in the `drift.js` file to our valid Drift key.
- [x] Invoke support via dropdown, see the Drift chat?
- [x] Invoke support via the left nav. see the Drift chat?
- [ ] Merge
- [ ] Deploy

NB: Unfortunately another branch (just called `drift`) went into `merge-branch` with these changes but I couldn't create a clean PR off of it because it was accidentally made from `merge-branch`. Could be after this is merged into `mainline`, `mainline` won't come back cleanly into `merge-branch` and it'll be a mess and we'll need to toss `merge-branch` and recreate it.


belucid:
@bago2k4 this looks good as is now. Ready for your review.

PR #669, opened 2018-10-08
Fix click out on iOS

bago2k4:
Card: 
Item:
> *When I tap the “•••” menu, and click somewhere else, it should close

iOS doesn't like window click listeners, they are never fired. I changed all those listeners to listen on `div#app` selector since it's a main div wrapping the whole app.

To test:
- check on desktop some click out (dismiss notifications, dismiss emoji picker, dismiss post contextual menu etc...)
- check on mobile dismiss of the post contextual menu

### DIFF
--- a/src/oc/web/components/reactions.cljs
+++ b/src/oc/web/components/reactions.cljs
@@ -34,7 +34,7 @@
                          (reset!
                           (::window-click s)
                           (events/listen
-                           js/window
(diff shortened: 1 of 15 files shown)


thepug:
Shouldn't this be a mixing or something reusable so that you don't have the listener defined in multiple places?   

Similar to https://github.com/kentor/react-click-outside

PR #686, opened 2018-10-16
Zindex fixes

bago2k4:
Bug: it's impossible to remove a user from a team right now since the confirmation dialog is hidden behind the settings panel.

I took this to fix also some other z-index rules that were arbitrary set and transformed them in relatively set. The base z-index is the navbar z-index (it was the first and most important we had at the beginning)

To test:
- move an activity (from the contextual menu, not cmail)
- share an activity
- edit an activity
- change the board of a post inside cmail
- add and remove a user from the team
- add and remove a user from a private section
- change a section settings: change slack channel, change name, save and close without saving
- repeat the editing ones on mobile

### DIFF
--- a/scss/partials/_activity_move.scss
+++ b/scss/partials/_activity_move.scss
@@ -1,7 +1,7 @@
 div.activity-move {
   width: 310px;
   height: 166px;
-  z-index: 1000;
+  z-index: 110;
 
   div.move-post-inner {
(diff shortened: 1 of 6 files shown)


bago2k4:
[Can't remove user from a team](https://trello.com/c/CnFMg34t/1574-cant-remove-user-from-a-team)

PR #688, opened 2018-10-16
Truncate org name

thepug:
This change truncates the org name to 127 chars.  The server side also truncates but this code prevents sending a long name to the server.

To test:

- sign up and create an org with a name longer than 127 chars
- [x] is the org created with a truncated name?


thepug:
[Max on org name length in Web UI and storage](https://trello.com/c/OnLmW6Q2/1565-max-on-org-name-length-in-web-ui-and-storage)

PR #689, opened 2018-10-17
Wall of difference

bago2k4:
Card: https://trello.com/c/8PhjLeVh

To test:
- add a post w/o videos or images but with a long body
- [x] do you see the body going from left to right when the stream is collapsed? Good
- add a post with an image and some text
- [x] do you see the image thumbnail on the right? Good
- [x] do you see the text only in the left part? Good
- expand it
- [x] do you see the text from left to right edge? Good
- [x] do you see the full size image now? Good
- add a post with a youtube video
- [x] do you see the video thumbnail on the right with a play button? Good
- [x] do you see the text only in the left part? Good
- expand it
- [x] do you see the text from left to right edge? Good
- [x] do you see the YouTube video now? Good
- add a post with a vimeo video
- [x] do you see only the play button on the right? Good (Vimeo thumbnails are currently broken)
- [x] do you see the text only in the left part? Good
- expand it
- [x] do you see the text from left to right edge? Good
- [x] do you see the Vimeo video now? Good

### DIFF
--- a/scss/partials/_stream_item.scss
+++ b/scss/partials/_stream_item.scss
@@ -62,7 +62,6 @@ div.stream-item {
     div.stream-header-head-author {
       float: left;
       width: 55%;
-      // overflow: hidden;
 
       @include activity-author(32);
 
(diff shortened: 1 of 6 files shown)


bago2k4:
[Wall of Difference!](https://trello.com/c/8PhjLeVh/1561-wall-of-difference)

PR #690, opened 2018-10-17
Gather slack channel data even if you aren't an admin.

thepug:
This change gathers slack channel info even if the team link isn't present. 

To test:
- sign up with slack as user 1
- add the slack bot as user 1
- sign up with same slack org as user 2
- [x] when sharing a post do you see the slack channels?


thepug:
[[Launch user issue] No channels for Slack share](https://trello.com/c/Ry8YAs4z/1576-launch-user-issue-no-channels-for-share)

PR #691, opened 2018-10-18
Org name length ui

bago2k4:
Set maximum length for the org name fields to 64 characters.
Also make sure the UI doesn't break if the org name is very long.

To test:
- signup
- insert the longest possible org name
- [x] is that 64 chars? Good
- finish onboard
- [x] check the navbar, org name is truncated and doesn't go in front of the search field? Good
- [x] check the user menu, org name and slug are truncated? Good
- [x] check the org settings panel, org name and slug are truncated? Good
- change the org name
- [x] can you NOT add an org name longer then 64 chars from settings panel? Good
- [x] save did work? Good

### DIFF
--- a/scss/mixins/_org.scss
+++ b/scss/mixins/_org.scss
@@ -75,6 +75,7 @@
       text-overflow: ellipsis;
       overflow: hidden;
       width: 100%;
+      max-width: 250px;
 
       @include mobile() {
         display: none;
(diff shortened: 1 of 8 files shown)


bago2k4:
[Max on org name length in Web UI](https://trello.com/c/hyR3Ncys/1585-max-on-org-name-length-in-web-ui)

PR #695, opened 2018-10-20
Fix mobile thumbnails

bago2k4:
BUG: thumbnails broke the mobile experience

Fix is complicated by the ziggeo videos that are showing the same size thumbnail on big web but they are shown full size on mobile.

To test:
- on mobile, check a stream of posts with:
- [ ] only headline and body
- [ ]  ziggeo video: should be shown full width and needed height
- [ ] body with at least one image and some text
- [ ] body with at a youtube or vimeo video and some text
- on desktop, check a stream of posts with:
- [ ] only headline and body
- [ ]  ziggeo video: should be shown as thumbnail and playable inline
- [ ] body with at least one image and some text
- [ ] body with at a youtube or vimeo video and some text

### DIFF
--- a/scss/partials/_stream_item.scss
+++ b/scss/partials/_stream_item.scss
@@ -171,6 +171,10 @@ div.stream-item {
           width: calc(100vw + 32px);
           height: auto;
           margin-left: -16px;
+          position: initial;
+          top: unset;
+          right: unset;
+          transform: translateY(0%);
 
           video {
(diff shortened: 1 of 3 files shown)


bago2k4:
[Missing org logo for Slack onboards](https://trello.com/c/o11kXX7K/1598-missing-org-logo-for-slack-onboards)

PR #696, opened 2018-10-22
FS url into drift user

bago2k4:
Until FS is out of data this won't work, also this may never work but won't give us any error.

To see the FS url you probably need to click on the user name on the right while in the Drift chat, then on the left sidebar click on See all attributes and look for `fsUrl` attr.


bago2k4:
Closing since this won't work if we don't upgrade

PR #700, opened 2018-10-22
Lose Made with Carrot from secure activity

bago2k4:
To test:
- open a secure activity (logout or use an incognito window)
- [ ] do you see the Made with carrot footer always visible? no need to scroll? Good
- click it
- [ ] are you redirected to the homepage? Good


bago2k4:
[Remove Made with Carrot modal from post obscurl](https://trello.com/c/3VbIs7HF/1599-remove-made-with-carrot-modal-from-post-obscurl)

PR #702, opened 2018-10-23
New & edit parameters

bago2k4:
Card: https://trello.com/c/c89sqgE8

Added 2 different parameters for the org:
- `new`: can be empty or have a board slug as value, open Cmail with the last used board or the specified board slug
- `headline`: is accepted only with `new` parameter and is used to specify a headline
- `edit`: need to be a valid activity uuid, open Cmail with the specified entry in edit mode

To test:
- signup to an org
- [x] visit AP, all good? Great
- now visite the same url adding `?new`
- [x] do you get Cmail to open with the last used section already selected? Good
- now visit AP with `?new=general` or another board-slug different than the one above
- [x] do you get Cmail to open with the specified board selected? Good
- click the delete button (to make sure the activity is not cached)
- now visit AP with `?new=general&headline=Monthlt%20update`
- [x] do you get Cmail to open with the specified headline already in place? Good
- now get the UUID of an existing activity of the current org
- visit: `?edit=activity-uuid` where activity-uuid is the id you just found
- [x] do you get Cmail to open with the specified activity in edit mode? Good
- now visit `?edit=1234-1234-1234`
- [x] do you NOT get Cmail to open? Good
- merge
- deploy

### DIFF
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -759,7 +759,6 @@
   (let [board-data (if (seq board-slug)
                     (dis/board-data (router/current-org-slug) board-slug)
                     (dis/board-data))]
-    (js/console.log "DBG get-board-for-edit board-slug" board-slug "board-data" board-data)
     (if (or (not board-data)
             (= (:slug board-data) utils/default-drafts-board-slug))
       (get-default-section)
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -807,7 +807,11 @@
 (defn cmail-reopen? []
   (when (compare-and-set! cmail-reopen-only-one false true)
     (if (contains? (router/query-params) :new)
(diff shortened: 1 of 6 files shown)


bago2k4:
[Create a new post CTA added to the bottom of specific transactional emails and the briefing](https://trello.com/c/c89sqgE8/1553-create-a-new-post-cta-added-to-the-bottom-of-specific-transactional-emails-and-the-briefing)

thepug:
When adding the headline the draft isn't saved when you click 'Save & Close'

bago2k4:
Hmmm not sure it's desirable that we save if the user never changed something... As soon as the user change something the post will be automatically saved, but saving the default title it's not something we want.
IE: i get an email reminding me to add a weekly update (we don't have this now, just prevision of what we could use this for), i click it and open carrot, i check but i don't have something to say so i close Cmail, why should we save that draft?
@thepug the debug statement was removed

PR #703, opened 2018-10-23
Ziggeo time of 5 minutes

thepug:
Limit the time of each video recording. 

To test:

- start a post and click record video
- [x] after clicking the start button do you see the ` 0:00 / 5:00` display in the top, middle?
- [x] does the recording stop after 5 minutes?


thepug:
[Limit video recording to 5 mins](https://trello.com/c/KfNz6KgD/1506-limit-video-recording-to-5-mins)

src/oc/web/components/ui/ziggeo.cljs
```diff
@@ -23,6 +23,7 @@
                                               :attrs #js {:width width
                                                           :height height
                                                           :autoplay autoplay
+                                                          :timelimit 300
```
bago2k4: @thepug why do you think we need the limit on the player?

PR #706, opened 2018-10-25
Mention with only email

bago2k4:
Card: https://trello.com/c/ySPKgxEa

If you have an active user with only email address it is going to break mentions.

Check this loom if you don't believe me https://www.useloom.com/share/4a28ac20409d4ff5adbd2897e02ebc61.

You can get an active user with only the email by inviting them and accepting the invitation but leaving the invitation flow before entering any data about yourself (you can enter a password if you need to come back).

To test:
- try the flow above and see that it breaks the mentions
- switch to this branch
- mention the above user with only email
- [x] can you mention him? Good

### DIFF
--- a/resources/public/lib/MediumEditorExtensions/MediumEditorTCMention/CustomizedTagComponent.js
+++ b/resources/public/lib/MediumEditorExtensions/MediumEditorTCMention/CustomizedTagComponent.js
@@ -32,7 +32,19 @@ function getUserDisplayName(user) {
 }
 
 function getUserSelectedDisplayValue(user) {
-  return (user["selectedKey"] === "slack-username")? user["slack-username"] : user["name"] || (user["first-name"] + " " + user["last-name"]);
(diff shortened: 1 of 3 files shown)


bago2k4:
[Mention empty user (with only email)](https://trello.com/c/ySPKgxEa/1612-mention-empty-user-with-only-email)

resources/public/lib/MediumEditorExtensions/MediumEditorTCMention/CustomizedTagComponent.js
```diff
@@ -146,6 +158,7 @@ class CustomizedTagComponent extends React.PureComponent {
 
   selectItem(user){
     let selectedValue = getUserSelectedDisplayValue(user);
+    console.log("DBG CustomizedTagComponent.selectItem", user, selectedValue);
```
thepug: remove this `console.log`

PR #710, opened 2018-10-27
Wrt improvements

bago2k4:
Cards:
- https://trello.com/c/w8oYmlVH
- https://trello.com/c/uq7ReqQl

### New logic for WRT

Added some more reliable ways to send the WRT read message, these are:
- reaction picked from picker
- reaction toggled
- comment add
- comment delete
- comment update
- comment reaction toggle
- Show more button is clicked
- post thumbnail or body is clicked
- Add a comment CTA is clicked
- Ziggeo video is played (small or big)
- post not truncated is scrolled in stream view

I also changed the function to detect the visibility to set read and seen messages, they now use the same function applied to different elements:
- seen: check the headline is visible
- read: check the not truncated body is visible

Visible means: the el is inside the viewport excluded the navbar, ie: (offset top of the element is below the navbar height AND above the bottom of the screen) OR (bottom of element is below the navbar height AND above the bottom of the screen).

To fix the remaining issue that happened to Stuart on Safari during the call we had on friday i applied a longer timeout to the `did-remount` check. It looks like the render cycle takes more time on a `did-remount` than on `did-mount`, that's why the check was failing for the newly added items.

### Testing

Since we did an extensive test on our Friday call i'll focus more on the remaining issue we had instead of testing all the above scenario (comment add, update, delete, react picked or toggled etc..)

To test:
- open Chrome with user A
- open Safari with user B
- nav to a new section
- have both window visible but focus only on user A/Chrome
- add an item with user A with short body and no media items (ie: body is not collapsed)
- [x] do you see the read count turning to 2 right after the item is added? Good
- now add another item
- [x] do you see the read count turning to 2 again? Good
- do it again for a few tiems
- now add a post with an image inside
- [x] do you see count to only 1 that is for user A? Good
- go to Safari and expand the post
- [x] do you see the count turn to 2? Good
- now with user A add a post with a Ziggeo video 
- [x] count 1? Good
- with user B play the video inline (no expand, just play)
- [x] count goes to 2? Good
- with user A add a long body post (at least 4 lines)
- [x] count is 1? Good
- with user B expand the post
- [ ] count increments to 2? Good
- let the sessions sit for at least 5 minutes (the more you wait the best it is)
- now w/o focusing back to Safari go to user A/Chrome and add a short body item
- [x] does the count turn to 2? Good


bago2k4:
[WST Socket Longevity](https://trello.com/c/uq7ReqQl/1626-wst-socket-longevity)

PR #714, opened 2018-10-31
Tweak What's New

belucid:
This takes advantage of tweaks from Headway that keep the UI from opening upward (there's never enough room in our UI to do that) and provides our own label for the show more link.

To test. 

- [x] Verify you see the new label text in the dialog.
- [x]  Make your screen very short, then trigger what's new in the drop down menu, does it not open upward cut off by the top of the browser? Good!
- [x] Use on mobile (deployed to staging)


belucid:
```
@include mobile() {
  div#HW_frame_cont {
    position: fixed;
    top: #{$navbar_height + 1}px;
    left: 0;
    width: 100vw;
    height: calc(100vh - #{$navbar_height + 1}px);
    z-index: #{$navbar_zindex + 200};
  }
}
```

belucid:
Ready to go again @thepug

PR #716, opened 2018-10-31
WRT viewers update

bago2k4:
Card: https://trello.com/c/ZV3esqy9
Abstract: https://share.goabstract.com/0be83571-fa7e-4848-8ef2-0c55a70c3758
(it's not the correct abstract link since that's about archiving but it has these changes too)

Add icon before WRT label (see design), change views to viewers.

To test:
- [x] see the person icon? Good
- [x] also on FF? Good
- [x] also on mobile? Good
- [x] does it say `x Viewer(s)`? Good


bago2k4:
[WRT viewers update](https://trello.com/c/ZV3esqy9/1651-wrt-viewers-update)

PR #720, opened 2018-11-01
In-app Only notifications

bago2k4:
Card: https://trello.com/c/SuIKZlvY

Change Never to In-app Only in notification frequency.

To test:
- go to Notifications settings
- change it to In-app Only
- save
- refresh
- open Notifications settings
- [x] do you see In-app Only? Good
- change it back no Email or Slack
- refresh
- open Notifications settings
- [x] do you see the last value set? Good


bago2k4:
["In-app only" Notifications](https://trello.com/c/SuIKZlvY/1658-in-app-only-notifications)

PR #726, opened 2018-11-01
Fix face pile

bago2k4:
Card: https://trello.com/c/MinnII7g

Face pile is off by one pixel (check the image attached)

To test:
- check that face pile is ok when it has 1, 2 and 3 (at least)


bago2k4:
[Fix face pile](https://trello.com/c/MinnII7g/1664-fix-face-pile)

PR #731, opened 2018-11-02
Fix cursor position in FF

bago2k4:
Card: https://trello.com/c/6jmspqZX

To test:
- in FF click + to add a new section
- [x] is the cursor in the right place? Good
- change the settings to private section
- focus the note field
- [x] is the cursor in the right place? Good


bago2k4:
[FF fix input fields](https://trello.com/c/6jmspqZX/1542-ff-fix-input-fields)

PR #739, opened 2018-11-13
Ws monitor

bago2k4:
Card: https://trello.com/c/f3mQHFLn

Needs this https://github.com/belucid/open-company-infrastructure/pull/89 for deploy

This has been sitting on staging for a while now, so not much testing is needed. You can check this locally or on staging with the following test:
- start web with this branch
- stop one of the following services: notify, interaction or change
- wait a random time
- restart the service
- go back to web console and in the error logs from Sente find the data `:udt-next-reconnect`
- check what time it says with: `new Data(:udt-next-reconnect)`
- wait to see if the service will reconnect
- also if you wait more than 2 hours before reconnecting and w/o touching that session you should see also a token refresh being made by the service client before reconnecting and exchanging the token (only notify and interaction, change doesn't use jwt)
- now you can check the sentry for the reports of send on a closed connection and/or connection down (every 30 sec, will be 1 minute on prod since it's an infra var)


bago2k4:
[Socket connection monitor](https://trello.com/c/f3mQHFLn/1669-socket-connection-monitor)

bago2k4:
[Socket connection monitor](https://trello.com/c/f3mQHFLn/1669-socket-connection-monitor)

thepug:
@bago2k4 some conflicts to resolve here. I'm not certain what to change in the user actions.

bago2k4:
@thepug fixed, also cleaned a bit more how we identify the user with Drift: i simply moved the code to a function in `oc.web.lib.chat` namespace.

src/oc/web/api.cljs
```diff
@@ -169,7 +169,8 @@
                         :path path
                         :method (method-name method)
                         :jwt (j/jwt)
-                        :params params}]
+                        :params params
+                        :sessionURL (when js/LogRocket (.-sessionURL js/LogRocket))}]
```
thepug: log rocket code needs to be removed

src/oc/web/api.cljs
```diff
@@ -193,7 +194,8 @@
 (defn- handle-missing-link [callee-name link callback & parameters]
   (timbre/error "Hanling missing link:" callee-name ":" link)
   (sentry/set-extra-context! (merge {:callee callee-name
-                                     :link link}
+                                     :link link
+                                     :sessionURL (when js/LogRocket (.-sessionURL js/LogRocket))}
```
thepug: more log rocket

src/oc/web/components/ui/login_overlay.cljs
```diff
@@ -305,20 +305,20 @@
...
+        (= (drv/react s dis/show-login-overlay-key) :signup-with-slack))
     (do
       (utils/after 150 #(router/nav! oc-urls/sign-up))
       [:div])
     ; password reset
-    (= (drv/react s user-store/show-login-overlay-key) :password-reset)
+    (= (drv/react s dis/show-login-overlay-key) :password-reset)
     (password-reset)
     ; form to insert a new password
-    (= (drv/react s user-store/show-login-overlay-key) :collect-password)
```
thepug: We should be consistent here.  Do we want to have the stores determine the method of retrieving information from the app state or use one large dispatch namespace to do that. I think having them in the store is the best.

thepug:
also conflicts

src/oc/web/components/ui/login_overlay.cljs
bago2k4: Well if we want to be consistent we should keep the keys in dispatcher, they are all there, that's why i moved this too.

We are keeping all the keys in dispatcher to avoid pulling stores to all the files the need that key or another. You only need to pull one file (dispatcher) to have the keys available.
Then the setter are in the stores.

I know that ideally is better to have the keys in the stores to avoid having a huge namespace but since some keys are using in more stores it's not very easy to change it now.
This is an example of it: the `show-login-overlay-key` is needed not only by the user store but also from the jwt store (that i created in this PR to pull out the JWT stuff to make it callable from the ws clients namespaces), if we keep it in the user store i need to pull it in the jwt store that is creating a circular dependency cycle.

I know you changed this but this is the only file that is keeping the keys inside, so: or we refactor all or we keep it consistent like it is right now.

I think at this point is cleaner this way.

PR #748, opened 2018-11-28
Fix Fullstory checks

bago2k4:
Card: https://trello.com/c/oBupmvDY

Issue: https://sentry.io/opencompany/oc-staging-web/issues/781088497/?referrer=slack

Improve the checks before calling Fullstory session URL function to avoid errors like the one above.

To test:
- run on localhost (don't change FS setup)
- have a service return a 400 or a 422 (or you can add any other error status at `oc.web.api:167`)
- now make sure the UI is doing that request
- [x] do you see a sentry reporting that error? Good
- [x] do you NOT see an error like the one above about `FS` not defined? Good
- merge

### DIFF
--- a/src/oc/web/api.cljs
+++ b/src/oc/web/api.cljs
@@ -170,7 +170,7 @@
                         :method (method-name method)
                         :jwt (j/jwt)
                         :params params
-                        :sessionURL (when js/FS (.-getCurrentSessionURL js/FS))}]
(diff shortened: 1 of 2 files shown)


bago2k4:
[Fix Fullstory checks](https://trello.com/c/oBupmvDY/1726-fix-fullstory-checks)

PR #755, opened 2018-12-03
WS authentication failed: closed

bago2k4:
Card: https://trello.com/c/FjI2y4nq

#### Bug
After adding more info to the failure of ws authentication i found out that sometimes the jwt is refused because the connection is closed.
Check this sentry https://sentry.io/opencompany/oc-beta-web/issues/779441312/?referrer=slack, if you scroll down to the additional data section you see under `rep` it says `closed`. That's the response we get for the auth request.
It probably means the connection was closed while it was attempting the auth.

#### Solution
So i added some code to handle the closed connection and retrying to connect.

#### To test:
This is hard to test since it's mostly a race condition, just check the code and check if you can get interaction and notify to connect locally and on staging.

### DIFF
--- a/src/oc/web/ws/notify_client.cljs
+++ b/src/oc/web/ws/notify_client.cljs
@@ -32,7 +32,7 @@
 (defn- send! [chsk-send! & args]
   (if @chsk-send!
     (apply @chsk-send! args)
-    (ws-utils/sentry-report "Notify" ch-state (first (first args)))))
+    (ws-utils/sentry-report "Notify" chsk-send! ch-state (first (first args)))))
 
 (defn notifications-watch []
   (timbre/debug "Watching notifications.")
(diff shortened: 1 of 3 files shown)


src/oc/web/ws/notify_client.cljs
```diff
@@ -59,6 +60,17 @@
                      (not (:udt-next-reconnect @@ch-state)))
             (utils/after (* 10 1000)
              (reconnect @last-ws-link (j/user-id)))))
+        (or (= rep :chsk/closed)
```
thepug: Could we clean up some of the code repeat here. Maybe a shared disconnect handler between the web socket clients?

bago2k4:
[Further investigation on websocket invalid JWT](https://trello.com/c/FjI2y4nq/1731-further-investigation-on-websocket-invalid-jwt)

bago2k4:
@thepug should be good now... it wasn't very clean to reused code since i had to pass in a lot of vars and also some functions to the common namespace.

Problem is that the clients namespaces have those vars and also the utility namespace is causing a circular dependency if i directly call the jwt actions in it.

Hope it makes sense, ping me with questions if you need

PR #756, opened 2018-12-03
Fix get board

bago2k4:
Card: https://trello.com/c/Lc2aXQhO

Error: https://sentry.io/opencompany/oc-beta-web/issues/790242397/?referrer=slack

I finally found the source of these get-board without link (maybe not all of them, but the one that was caused by this trace at least).

Solution: i found out that when we needed to refresh the current data we were checking if user was on all posts, if so we were reloading the AP data, if not we were trying to refresh the current board. If user was on Must See  we were looking for a board with slug must-see in the org data, this was leading to a get-board without link error.

To fix this i also had to do some cleanup that included removing the `:from-all-posts` key from the router that was used to track user location before opening a fullscreen post or the edit view (before cmail since it was a fullscreen view), this lead to remove a couple of unused components too.

To test (try with and without this change):
- go to Must See
- add a post
- edit it
- publish it
- [x] did it work? Good
- [x] no errors? Good


bago2k4:
[Get board without link web error](https://trello.com/c/Lc2aXQhO/1733-get-board-without-link-web-error)

PR #766, opened 2019-01-02
Validate that the id token has the requested secure uuid.

thepug:
Test using: https://github.com/open-company/open-company-storage/pull/181

To test: 

- generate a valid id token url from digest
- logout of carrot
- open the url from digest
- [x] did it work as before?

- change the uuid in the url (`/post/<uuid>`)
- [x] is the 404 page displayed?


thepug:
[Identify post uuid when using id token](https://trello.com/c/9c5nMKyG/1769-identify-post-uuid-when-using-id-token)

PR #767, opened 2019-01-04
Check id token for authenticated post url.

thepug:
Uses the secure uuid and component. However this is gathered from the token and not leaked in the url. 

Test steps:

- generate a url from the digest service (email)
- [x] does the url have the id token and is it the private (not secure uuid) url?
- [x] does the url work if the user is logged in?
- [x] does the url work if the user is logged out?

- generate a url from the digest service (Slack)
- [x] does the url have the id token and is it the private (not secure uuid) url?
- [x] does the url work if the user is logged in?
- [x] does the url work if the user is logged out?

- generate a notification by mentioning someone
- repeat the steps above with the email and bot service to check the generated links


thepug:
[ID Token issues](https://trello.com/c/rd0G1cVF/1759-id-token-issues)

PR #768, opened 2019-01-07
Extra settings

bago2k4:
Card: https://trello.com/c/JcvNr1Ew

Adds separate settings for digest, notifications and reminders.

Supports open-company/open-company-digest#13

Test instructions there.


thepug:
http://recordit.co/cQBIXNBggz  This user was email only (no slack account)  So they won’t get the notifications/digest/reminders at all if they select slack.

thepug:
In this video I have just signed in with slack and clicked `Notifications`.  http://recordit.co/RqnUhjAWwh
It says `Add a team`.  I would think it would enable bot since there is an existing team.

belucid:
> http://recordit.co/cQBIXNBggz This user was email only (no slack account) So they won’t get the notifications/digest/reminders at all if they select slack.

Good catch @thepug 

This one is for you @bago2k4

bago2k4:
@thepug fixed both things...
I suggest you test also the above case with users that are contributors and not admin. To make sure the Slack orgs and team data are available (contributors don't have access to the team data so they can't add/remove bot, also they can't know if there is a bot already installed).

Good catch!

PR #771, opened 2019-01-14
Test for current activity id to allow share urls to work.

thepug:
Fix for shared urls that are available to all.

To test:
copy a public/shared url 
- [x]  when opening the url while logged in does it redirect to the private url?
- [x] when opening the url while logged out ,does it show the post but asks you to login?

- [x] open a digest url while logged in, does it open the private url?
- [x] open a digest url while logged out, does it open the url as the user but asks the user to login?
-  change the digest url to a different post uuid
- [x] does it 404?

### DIFF
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -509,8 +509,11 @@
    (some #{status} [401 404])
    (router/redirect-404!)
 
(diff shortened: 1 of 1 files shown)


thepug:
[Auto-share links are broken](https://trello.com/c/iPZ440uE/1789-auto-share-links-are-broken)

PR #772, opened 2019-01-14
Queue websocket messages when connection is down. 

thepug:
This change will save all data sent to a web socket if that web socket is closed.  When it returns the saved data will be sent to the server.

To test:
This is meant to resolve the id token mark as read , so we will test that.

- generate a digest url with an id token
- logout of carrot
- open to digest url in your browser
- [ ] does the console have the following?:
```
DEBUG [oc.web.ws.change-client:60] - Sending item/seen for container: 8a40-4df7-8df4 item: 898d-42ab-a06b
raven.inc.js:58 DEBUG [oc.web.ws.change-client:64] - Sending item/read for container: 8a40-4df7-8df4 item: 898d-42ab-a06b
raven.inc.js:58 DEBUG [oc.web.ws.change-client:160] - Push event from server:  [:container/status ()]
raven.inc.js:58 DEBUG [oc.web.ws.change-client:93] - event-handler :container/status
raven.inc.js:58 DEBUG [oc.web.ws.change-client:106] - Status event: ()
raven.inc.js:58 DEBUG [oc.web.ws.change-client:160] - Push event from server:  [:container/status ({:unseen [], :container-id "8a40-4df7-8df4"})]
raven.inc.js:58 DEBUG [oc.web.ws.change-client:93] - event-handler :container/status
raven.inc.js:58 DEBUG [oc.web.ws.change-client:106] - Status event: ({:unseen [], :container-id "8a40-4df7-8df4"})
```

- [ ] in the change service logs is there a mark as read/seen event?
something like 
```
[websocket] item/read for: 898d-42ab-a06b in: 8a40-4df7-8df4 at: 2019-01-15T15:26:54.459Z by: 96b3-4f31-9fe1 / 16f1611e-b28a-417d-8854-1312ecf49d03
```

- check that normal web socket operations work with the change service 
  Different users see changes.  Things are marked as read
- check that normal operations work with notify service
Notifications are received  in client
- check that normal operations work with the interaction service.
Different users see reactions in real time.


thepug:
[Race condition on watch](https://trello.com/c/IA4Prcxv/1247-race-condition-on-watch)

src/oc/web/ws/utils.cljs
```diff
@@ -6,9 +6,39 @@
...
+  (when (and chsk-send! (:open? @@ch-state))
+    (doseq [qargs @cmd-queue]
+      (apply chsk-send! qargs))
+    (reset-queue)))
+
+(declare sentry-report)
+(defn not-connected [service-name chsk-send! ch-state & args]
+  (let [arg (first (first args))]
+    (buffer-cmd arg)
+    (sentry-report service-name chsk-send! ch-state arg)))
```
bago2k4: @thepug i think we can remove this, WYT?
Before it was useful to know how many messages were sent over a closed socket, because they were lost messages. Now we queue them up, so if you see another place were we could be losing messages (for example if when you go to drain the queue the socket is closed again and you lose them) then we should add a sentry report there.
Let's discuss in #development

bago2k4:
@thepug simplest way to test this (works only on localhost):
- stop interaction and change service
- refresh web client
- check how many messages you have in the console with `cljs.core.deref(oc.web.ws.utils.cmd_queue)`
- restart interaction service only
- check in Network tab of dev tools what messages are sent to interaction service when connection come back up

PS: if service is down Sente already handles the reconnect, but it uses an exponential wait time algorithm so the more it waits to reconnect the more it will take to retry, you can see when the next reconnect will happen looking at the `:udt-next-reconnect` parameter of these errors: `Send over closed {service-name} WS connection`.

thepug:
@bago2k4 Lets remove the sentry stuff later

PR #787, opened 2019-01-29
User timezone

bago2k4:
Card: https://trello.com/c/RZeMLiaa

This card is intended to make sure every user has timezone set, even if they don't set it in their profile.

To test:
- signup via email
- [x] once in carrot check that the email user has timezone set
- invite someone via email
- [x] once in carrot check that the invited user has timezone set
- signup via Slack
- [x] once in carrot check that the Slack user has timezone set
- invite a user via Slack
- [x] once in carrot check that the Slack invited user has timezone set
- signup via google
- [x] once in carrot check that the google user has timezone set


bago2k4:
[Make sure users have TZ set](https://trello.com/c/RZeMLiaa/1814-make-sure-users-have-tz-set)

PR #799, opened 2019-02-20
UX desktop: share tweaks

bago2k4:
Card: https://trello.com/c/B0D7rg0d

Remove share via email, tweak URL and Slack share style

To test:
- use an org w/o Slack
- share an item
- [x] check that the share dialog style is ok w/o the tabs row
- [x] can you share via URL only? Good
- [x] can you switch btw public and team url type? Good
- copy public url
- open in incognito window
- [x] do you see the post as anonymous in the secure post view? Good
- copy the team url
- open in incognito window
- [x] do you see login wall? Good
- open it in you logged in session
- [x] do you see the post now? Good
- now use an org with Slack bot
- share an item via Slack
- pick a channel
- add a note
- share
- [x] did it work? Good


bago2k4:
[UX Desktop: Share tweaks](https://trello.com/c/B0D7rg0d/1872-ux-desktop-share-tweaks)

PR #805, opened 2019-02-21
Ux desktop mention notifications

bago2k4:
Card: https://trello.com/c/PK2ASsnE

To test:
- org with user A and B both contributors or admins
- add a post with user A
- with user B comment on the post mentioning user A and with a link
- [x] does the link in the add comment box and then in the list of comments is green? Good
- [x] does the notification in user A session looks like the new design? Good
- [x] does the link is green in the notification and in the notifications tray? Good


bago2k4:
[UX desktop: mention notifications update](https://trello.com/c/PK2ASsnE/1879-ux-desktop-mention-notifications-update)

PR #806, opened 2019-02-21
UX desktop: comments scroll in stream view

bago2k4:
Card: https://trello.com/c/48mgWviZ

To test:
- add a short post
- click on add a comment
- [x] did it scroll to comments? Good
- comment on the post
- collapse the post
- click on the comments count
- [x] did it scroll to the comments? Good
- now add a long comment
- [x] repeat the comments scroll test above
- on mobile
- [x] does it scroll to the comments when you click the comments count? Good


bago2k4:
[UX desktop: comments scroll in stream view](https://trello.com/c/48mgWviZ/1882-ux-desktop-comments-scroll-in-stream-view)

belucid:
@bago2k4 when I click "Add a comment" then "Cancel" it leaves the comment text box. Shouldn't that go away?

bago2k4:
@belucid no the add comment box is always visible when the post is expanded, or you wouldn't be able to add a new comment anymore.
The only things that go away are the buttons to submit the comment. Makes sense?

belucid:
@bago2k4 that makes sense thanks.

Seems there's another issue though.

Sequence:

- On a long collapsed post
- Click "Add a comment"
- Click "Show less"

Notice that "Show less" acted like "Cancel". The "Show less" is still there and you have to click it again.

bago2k4:
@belucid yeah i had that item in the UX FoC card but we diceded to remove it since it's a side effect of how it works now.

I explained it to Stuart this morning in #general and he agreed on delaying this.

Explanation:
when you click the show less button the first event that is triggered is the unfocus of the add comment box, that causes the cancel/send buttons to disappear and the Show less button to move up (this is all by design). All this happens on mouse down event, when you release the click (on mouse up) the button is not anymore under your mouse so it’s not interpreted as a click on that button.

bago2k4:
Fixing that imply adding some complexity to those events that now have the most natural flow: now we use blur for the add comment and click for the show less button, i should change one or both those events (ie: simulate blur with on window click or use on mouse down instead of on click for the show less event) and that could lead to weird situations and/or bugs

belucid:
@bago2k4 2nd test in the test steps didn't work

Also for the 1st test, should scroll so the green comment button is in view, otherwise you don't know it's there.

bago2k4:
@belucid i am not sure it should scroll to the green comment button. In case you have 0 comments it says: Add a comment so that would be good. But if you have 10 comments it says `10 comments` so if you click it you expect to see the comments, not the add comment box.
So i think we should scroll to the first comment in the list.
If you want to add a comment you need to scroll down yourself.

bago2k4:
@belucid i tested again on Chrome and FF and they both scrolls, maybe you were already scrolled to the needed point?
not sure what happened, can you provide a loom if you keep seeing the issue?

belucid:
@bago2k4 y, for the green button I was specifically referring to the "Add a comment" case.

bago2k4:
@belucid in that case the add comment is always the first thing under the reactions/comments summary line.
To calculate the scroll position i take the top of the reactions and subtract 200 px (Stuart asked to show a piece of the post too). So the add comment is always in visible in that case

PR #256, opened 2017-09-12
UI/UX Polish

bago2k4:
Cards:
- https://trello.com/c/a9hdPwjk
- https://trello.com/c/2bMHClM9
- https://trello.com/c/xTFUUC1a

Inherited cards from closed PRs (with branch name):
-  `stories` https://trello.com/c/kHqOZct2
- `team-settings` https://trello.com/c/P50o4uj8
- `onboarding-rebranding` https://trello.com/c/9APZ6D5O

Issue from `stories` branch PR:
- [x] Open a new update. Add a comment. Add a 2nd comment. Close the update to go back to the board. Notice the board says "1 comment" rather than 2.

PR #807, opened 2019-02-21
UX desktop: comments count fix

bago2k4:
Card: https://trello.com/c/OzDxeUwY

To test:
- add a comment
- add another
- now delete one
- [x] comments count at the top of the comments is right? Good

Regression test (we saw in #256 ):
- go to the post view (copy url from share)
- add comment
- add another
- delete a comment
- go back to AP
- [x] comments count is right? Good


bago2k4:
[UX desktop: comments count](https://trello.com/c/OzDxeUwY/1883-ux-desktop-comments-count)

belucid:
@bago2k4 the comment count is now accurate for deletes, but the facepile isn't. It keeps the avatar of deleted comments.

bago2k4:
@belucid right but the face pile was already working that way, i didn't touch at all that code, i think we can merge this and add another issue for the face pile to fix.
We have also to fix the uniqueness by avatar id for the repetition when commenting from Slack.
WYT?

belucid:
OK @bago2k4 added both of these to https://trello.com/c/pR6wLnZJ

PR #812, opened 2019-02-25
Boot check throw

bago2k4:
This is needed to have travis report errors from yagni, eastwood, kibit and bikeshed.
Also fixed all the warnings from eastwood.


bago2k4:
@belucid last but one commit above this comment was a test, i added back a problem for eastwood and kibit, travis failed, the last commit instead is where i removed it and travis succeeded.
I think we are good now

PR #814, opened 2019-02-26
Debounce scroll

bago2k4:
Debounce scroll function, for the moment applied only to the seen/read mixins.

To test:
- with user A add some new posts (enough to scroll out of the way)
- with user B open all-posts
- start scrolling down
- [x] with user A check that you get reads for all the posts
- [x] also check in the change console or socket that seen and read messages are sent for all the posts


belucid:
@bago2k4 what's the status of this one?

bago2k4:
@belucid i removed the debounce from the dashboard-layout component. It was used to show the + button when scrolling down. Problem is that it lose some scrolls steps with the debounce and can be that the last step is important to show/hide the button and it's missed.
The rest shouldn't be affected by this kind of issue, moving this in review

PR #817, opened 2019-02-27
UX mobile: new tag

bago2k4:
Card: https://trello.com/c/Lj1Kc25a

To test:
- check the NEW tag on the posts (you can simply replace `(:new activity-data)` with `true` in stream_item.cljs)
- [x] does it looks like design (has a yellow bell in it)? Good
- [x] is it aligned as it should? Good
- [x] check it on mobile too


bago2k4:
[UX mobile: new tag](https://trello.com/c/Lj1Kc25a/1897-ux-mobile-new-tag)

PR #824, opened 2019-03-06
Verify route fix

bago2k4:
Card: https://trello.com/c/r5FPce9o

Bug: when visiting /verify route without a token we show a blank screen and an error is reported to sentry.

Fix: redirect to user org or homepage (in case user is not logged in) when visiting /verify without a token.

To test:
- signup
- wait for the verification email
- click the link
- [x] did it work? Good
- [x] make sure you have verified email (open QSG for example)
- now visit /verify (without a token as get parameter)
- [ ] did it redirect you back to your org? Good
- logout
- visit /verify
- [x] did it redirect you to homepage? Good


bago2k4:
[Verify route error](https://trello.com/c/r5FPce9o/1914-verify-route-error)

belucid:
@bago2k4 this doesn't work as the testing steps describe.

Accessing `http://localhost:3559/verify` logs me out.

bago2k4:
@belucid i moved the user logout after the check if the token exists.
That mean that you are redirected to your org if you are logged in now.
The problem though is another, not sure why but verify works only if user is logged out. So if i have the wrong link and i click it i am first logged out then i get the link not good page and now i am left out of the site.
I think we should fix this, WYT?
Meaning we should leave the user logged in if he is and check if the user he is verifying is the one he is actually logged in now, not sure what to do if the 2 don't match...
Backlog card?

PR #825, opened 2019-03-06
Ws reconnect improvements

bago2k4:
Card: https://trello.com/c/jtu4Gqvm

Testing:
it's not very easy to test this since Sente has automatic reconnect handling with growing wait time. This checks every minute that in case one of the WS services is disconnected, Sente has a next reconnect time set, if it's not present (no `(:udt-next-reconnect @@state)`) it will force a reconnect.
Also it should remove all the `Send over closed WS connection` that are not relevant since we have queues now.
The `Send over closed WS connections` that keep them until a connection is available.
To test just check that the WS services are working as expected and that the client reconnect after services are stopped and restarted.


bago2k4:
[Fix WS connection checks and sentry reports](https://trello.com/c/jtu4Gqvm/1913-fix-ws-connection-checks-and-sentry-reports)

PR #828, opened 2019-03-08
Multicompany chevron

bago2k4:
Card: https://trello.com/c/YxRyUnTU

To test:
- with a normal user go to AP
- [x] do you see you org name and logo up top? Good
- [x] do you NOT see the orgs dropdown chevron? Good
- now setup a user with multicompany
- logout and back in with the user if you haven't already
- [x] do you see the dropdown chevron even when you don't hover the org logo/name? Good


bago2k4:
[Multicompany chevron](https://trello.com/c/YxRyUnTU/1921-multicompany-chevron)

PR #830, opened 2019-03-08
Show less jumps fix

bago2k4:
Card: https://trello.com/c/yFOWlAZq

To test:
- add a long post
- expand the post
- focus on add comment
- click on Show less
- [x] does the post gets collapsed back? Good
- before it was only removing the focus to the add comment field causing a jump in the view, but no collapse

### DIFF
--- a/src/oc/web/components/stream_item.cljs
+++ b/src/oc/web/components/stream_item.cljs
@@ -290,7 +290,10 @@
             [:button.mlb-reset.expand-button
               {:class (when expanded? "expanded")
                :ref :expand-button
-               :on-click #(expand s (not expanded?))}
+               :on-click #(when-not expanded?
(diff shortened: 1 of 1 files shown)


bago2k4:
[Show less jump](https://trello.com/c/yFOWlAZq/1923-show-less-jump)

belucid:
@bago2k4 If I type a comment in, then click "Show less", my comment is gone when I expand again via Show more or "Add a comment...". That's not good.

bago2k4:
@belucid hmm y the comment is kept in the local component state, that is lost if you collapse the post.
This is unrelated to this though, i need a new PR to fix it.
Let me know if you add a trello card or want me to

bago2k4:
@belucid added a new card for it

PR #839, opened 2019-03-19
Fix page white band

bago2k4:
Card: https://trello.com/c/EG1bL939

BUG: if you have a lot of sections, go to an empty one and resize the browser window to not be tall enough to fit all the sections list you see a white band at the bottom if you scroll down.

This fixes the above issue for desktop by setting the left navigation bar to relative positioning.

To test:
- use an org with ~10 sections
- nav to an empty section
- with full height browser window (make sure the left navigation bar is completely visible and fixed, doesn't scroll with the page)
- [x] everything good? Good
- [x] invite, reminders and support buttons are at the bottom of the page? Good
- now resize the window so the left navigation bar has not enough space to be totally visible
- [x] while resizing do you NOT see the top of the navigation bar moving? Good
- [x] once resized, if you scroll down to the bottom of the page do you NOT see a white bad? Good
- now switch to mobile and refresh
- open the navigation menu
- [x] everything good? Good

### DIFF
--- a/scss/partials/_navigation_sidebar.scss
+++ b/scss/partials/_navigation_sidebar.scss
@@ -265,7 +265,7 @@ div.left-navigation-sidebar {
   }
 
   &.navigation-sidebar-overflow {
-    position: absolute;
+    position: relative;
     overflow-x: unset;
     overflow-y: unset;
     overflow: auto;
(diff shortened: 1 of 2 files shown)


bago2k4:
[Fix white band on small screens](https://trello.com/c/EG1bL939/1935-fix-white-band-on-small-screens)

PR #847, opened 2019-04-23
Fixed the regex for the email and domain validation in the html forms.

### LINKED ISSUE #846: Email format validator has false negatives on funky domain extensions
My email address was reported as incorrect because it finishes by `.taipei`.

`.tw` works, so I guess that there is a whitelist somewhere that may need an update - or maybe there shouldn't be a whitelist.

I spotted the problem in 2 places:

email-format-problem

image

However, the validation problem did not prevent me from progressing through the registrations.

green-coder:
Not ready to be merged yet, I still need to test it.

Fixes #846

### DIFF
--- a/src/oc/web/components/ui/onboard_wrapper.cljs
+++ b/src/oc/web/components/ui/onboard_wrapper.cljs
@@ -101,6 +101,7 @@
             {:type "email"
              :class (utils/class-set {:error (= (:error signup-with-email) 409)
                                       utils/hide-class true})
+             :pattern utils/valid-email-pattern
              :value @(::email s)
              :on-change #(let [v (.. % -target -value)]
                            (reset! (::password-error s) false)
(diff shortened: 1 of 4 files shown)


bago2k4:
@green-coder no rush, just remember to pull latest mainline in before requesting review. It has conflicts

green-coder:
I was able to run the `auth`, `storage`, `interaction` and  `web` services, and I tested the email matching on the account registration page. It seems to work fine.

I did not find yet how to register a user on my instance, so I did not test the page which tests only the domain part of the email. But I expect it to work as the code is very similar to the one I tested above.

bago2k4:
Hi vincent, thanks, don't worry about registration. This should be enough.
I'll take a look very soon. Sorry for the late reply but i have been out
since yesterday and I'm still on mobile today.

bago2k4:
This looks good @green-coder , merging in.
Thanks

PR #849, opened 2019-04-24
Carrot 2.0

bago2k4:
Card: https://trello.com/c/ISkq1g46

This is the first and (probably the only) giant PR for Carrot 2.0.

In this branch there is all the UI refresh done in the last few days plus the web side work of [this card](https://trello.com/c/tSMO0Qhf).

To test:
- [x] test the org settings panel
- [x] open org settings, stop storage, test org settings errors
- [x] test the integrations panel
- [x] test errors states for integrations (you should try faking the redirect back urls with access=false etc, let's discuss more if you need)
- [x] test the team management panel
- [x] test the team management panel errors (stopping auth service)
- [x] test the invites panel
- [x] test the invites panel errors (stopping auth service)
- [x] test user settings panel
- [x] test user settings panel errors (stop auth service)
- [x] test user notifications panel
- [x] test user notifications panel errors (stop auth service)
- [x] test reminders panel
- [x] test reminders panel errors (stop reminders service)
- [x] test All Posts: load more items, more menus, reactions
- [x] test section: scroll items, more menus, reactions
- [x] test section settings panel
- [x] test section settings panel errors (stop storage service)
- [x] test navigation sidebar
- [x] test the expanded post view: post with only text, with images, with videos, with attachments (1 or more)
- [x] test comments: add, edit, delete

Test read send:
- [x] Not when there is truncated body
- [x] Not when body contains at least one image
- [x] Not when body contains at least one video (Youtube or Vimeo)
- [x] Not when body contains a recorded (Ziggeo) video
- [x] Not when post has comments
- [x] Not when post has attachments
- [x] Yes when body is not truncated and all the above are not met

PR #850, opened 2019-04-25
Whats new dot

bago2k4:
#### NB: review after #849 

To test:
- open carrot
- wait a few seconds
- [x] do you NOT see a green dot besides your user avatar? Good
- add a new whats new item to headway
- refresh Carrot
- wait a few seconds
- [x] do you see a green dot besides the user menu? Good
- now open the menu
- [x] do you see a green dot besides what's new too? Good
- click Whats new
- [ ] do you see the headway widget in the right place? Good

NB: to make this work i removed the animation from the menu. Since that's almost always the first panel that opens i think it makes sense to remove all the panels animations at this point, WYT?

NB: this doesn't remove the page scroll when the menu (or probably another panel) appear, i won't fix it here though since that's fixed in #866

### DIFF
--- a/scss/partials/_edit_recurring_update_modal.scss
+++ b/scss/partials/_edit_recurring_update_modal.scss
@@ -11,16 +11,8 @@ div.edit-recurring-update-modal-container {
   text-transform: none;
   background-color: $carrot_modal_bg;
   z-index: #{$navbar_zindex + 156};
-  opacity: 0;
+  opacity: 1;
   transition: opacity 180ms ease-in;
-
-  &.appear {
-    opacity: 1;
-
-    div.edit-recurring-update-modal {
-      right: 0;
-    }
-  }
 
   button.modal-close-bt {
     position: absolute;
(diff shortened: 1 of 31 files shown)


belucid:
@bago2k4 seems the merge is still possibly bad? I don't get a clean app start in the browser on this branch at the moment (see Slack)

belucid:
@bago2k4 after clicking through what's new and seeing the Headway widget, the numbering badge on the what's new in the menu goes away, but the green dot on the avatar in the header does not, it stays there until there is a page refresh.

bago2k4:
@belucid in one of the feedback Ryan wanted the Whats new item to open a new tab pointing to https://whats-new.carrot.io/
If we do that though the badge count is not cleaned out, and i don't think i can force that since it's a cookie set on under that domain...
WYT?

belucid:
@bago2k4 let's leave it as-is for now. We need the badge to clear out.

bago2k4:
@belucid added a note in the description, repeating here to make sure: you shouldn't care if the page can't still scroll up and down when the menu (or another panel) is open, that will be fixed in #866

PR #853, opened 2019-04-29
Why it matters

bago2k4:
Card: https://trello.com/c/RAiNROTP

Related PRs:
- [Storage](https://github.com/open-company/open-company-storage/pull/195) *Has test steps*

To test:
- [x] review with design of Cmail
- [x] review with design of Cmail fullscreen
- [x] try adding a post with small editor
- [x] try adding a post with fullscreen editor
- [x] try to use all the small editor functionality
- [x] try to use all the fullscreen editor functionality
- [x] open a post with its public share link, do you see the abstract?
- [x] open a post with a digest link (with your id-token), do you see the abstract?

~Test read send:~

~- [ ] Not when there is truncated body~
~- [ ] Not when body contains at least one image~
~- [ ] Not when body contains at least one video (Youtube or Vimeo)~
~- [ ] Not when body contains a recorded (Ziggeo) video~
~- [ ] Not when post has comments~
~- [ ] Not when post has attachments~
~- [ ] Not when post has abstract (we only show it, not the body in FoC)~
~- [ ] Yes when body is not truncated and all the above are not met~

NB: This one keeps the footer with all the media picker buttons, that will change in a subsequent PR that will go back to the inline media picker for Medium Editor

NB2: i accidentally merged this into `public-post-refresh` to update the branch, but before i had merged that branch into this. So i guess now the only way is to review them together: https://github.com/open-company/open-company-web/pull/861


bago2k4:
[Why it matters/real fullscreen](https://trello.com/c/RAiNROTP/1995-why-it-matters-real-fullscreen)

belucid:
@bago2k4 are these test steps for read send out of date? Final decision was that read is only sent when full page is opened, or manual mark as read is used and never otherwise, even if everything you can see is already there FoC.

bago2k4:
Hmm can be, I'll review them monday morning. Thanks

bago2k4:
@belucid i strikethrough the read test steps, those steps are still actually real for this PR since the read send will change only in the `mark-as-read` branch, here i only adjusted the previous read code.

src/oc/web/lib/utils.cljs
```diff
@@ -636,6 +636,10 @@
 
 (def default-headline "Untitled post")
 
+(def default-abstract "Post preview: summarize why this post matters…")
+
+(def max-abstrct-length 280)
```
belucid: @bago2k4 this isn't used anywhere

src/oc/web/mixins/ui.cljs
```diff
@@ -241,4 +241,24 @@
     (do
       (events/unlistenByKey (:on-resize-listener s))
       (dissoc s :on-resize-listener))
-    s))})
\ No newline at end of file
+    s))})
+
+(defn autoresize-textarea [ref & [initially-focused]]
```
belucid: @bago2k4 docstring explaining what this one is all about would be good

src/oc/web/lib/utils.cljs
bago2k4: it used in `oc.web.components.cmail;569`

PR #856, opened 2019-04-29
Mark as read

bago2k4:
#### Review after #849 

Card: https://trello.com/c/KRA9KvKU

Related PRs:
- [Storage](https://github.com/open-company/open-company-storage/pull/193)
- [Change](https://github.com/open-company/open-company-change/pull/18)
- [Infra](https://github.com/belucid/open-company-infrastructure/pull/107)
- [Change 2](https://github.com/open-company/open-company-change/pull/20)

#### Review after #849 

To test:
- [ ] test must see posts are read when they are opened (not on FoC)
- [x] test you can mark a must see post as read/unread
- [ ] test normal posts are read when they are opened (not on FoC)
- [ ] test you can mark a normal post as read/unread


bago2k4:
[Ability to mark a post as unread](https://trello.com/c/KRA9KvKU/1997-ability-to-mark-a-post-as-unread)

src/oc/web/actions/activity.cljs
```diff
@@ -640,7 +644,15 @@
...
-            (all-posts-get (dis/org-data) (dis/ap-initial-at)))
+            (all-posts-get (dis/org-data) (dis/ap-initial-at)
+             ;; In case another user is adding a new post
+             ;; mark it as unread directly
+             (when (and (= change-type :add)
+                        (not= (:user-id data) (jwt/user-id)))
+               (fn [{:keys [success]}]
+                 (when success
+                   (dis/dispatch! [:mark-unread (router/current-org-slug) {:uuid activity-uuid
+                                                                           :board-uuid section-uuid}]))))))
```
cjsauer: @bago2k4 I'm having trouble understanding this section of code. Why is it necessary to explicitly mark a new post as unread when it comes from the change service? Shouldn't a new post already default to the unread state? I'm not so much criticizing the code as much as I am trying to understand it.

bago2k4: Good question indeed.
So the change service doesn't say what posts are read, instead it says which one are unread. Not sure this is the optimal way to do this but it was how the seen was working and we decided to use the same logic for read (for dev time and also logic complexity).
As a result of that if a new post is added from the current session that post will be in read state until Change Service receive the message (SQS message) that a new post has been added and push to the client that as an unread.
To avoid any possible delay introduced by SQS or WS connection I make sure the newly published post is marked as unread, only in the current local state though.
Make sense @cjsauer ?

cjsauer: @bago2k4 yes that makes sense, thanks for the explanation

bago2k4: @cjsauer yeah, i was actually referring to another situation where the post is added from the current user.
You were asking about a situation where the post is being added from another user and the client receives it from the CS connection.
But the logic is still the same, the post is marked directly as unread to avoid delays btw the new post add and the refresh of the list of unreads posts.

cjsauer:
@bago2k4 when testing with two separate users in two separate browsers, when one user clicks "Mark as Unread", the other user is not seeing the WST count updating. For example:

1. User A creates post. WST count is 1.
2. User B views post. WST count is 2.
3. User B marks post as unread. WST count is still 2, for both users.

bago2k4:
Reads are not pushed live @cjsauer , they are loaded when the app starts and refreshed when you open the WRT dialog.

cjsauer:
@bago2k4 okay. Opening the WST dialog doesn't appear to update the count either. Should we expect it to do that?

belucid:
@bago2k4 please resolve the conflicts here.

cjsauer:
@bago2k4 so it looks like the above issue was resolved, but I'm seeing another problem where the post author's stream view is highlighting the post that they just created. To reproduce:

1. Create a post with User A. This will redirect you to the expanded view.
2. Click the "Back to {Board}" button.
3. Observe that the FoC is still highlighted even tho the author has viewed the post.

cjsauer:
@bago2k4 I'm still noticing FoC highlighting issues.

1. Make a post with User A. You're taken to expanded view, and WST is correct. Click "Back to {Board}" and notice that the card is incorrectly highlighted as unread.
2. With User B, navigate to the board that you just posted to, and notice that the FoC is _not_ highlighted.

Basically the highlighting is opposite of what's intended.

cjsauer:
@bago2k4 still one case of FoC highlighting that seems to off:

1. With User A, navigate to Board X
2. With User B, navigate to Board Y
3. Make a post with User A in Board X
4. User B should see the Board X's name bolded in the nav menu. Click it, and notice that FoC highlighting is not displayed properly.

I noticed that if User B is already viewing the same board (Board X in this example) then FoC highlighting is correct. It's only when I _switch_ to Board X after a post has been made that highlighting is not correct.

PR #862, opened 2019-05-06
Foc hover fixes

bago2k4:
Card: https://trello.com/c/OonUscAM

To test:
- hover a read item on FoC
- [x] do you see both top and bottom border disappear into the bg color of the hovered tile?
- now add 2 must see items with another user
- hover one of them
- [x] do you see a box shadow around the hovered tile?
- [ ] respect design?


belucid:
@bago2k4 the design here seems a little incomplete and/or dated with unread hover, but we're addressing that in other work so we'll merge this for now.

PR #872, opened 2019-05-16
Fix FoC hover icons

bago2k4:
Card: https://trello.com/c/JORTuPyb

To test:
- add 2 new post
- add a comment to one
- go to All Posts
- [ ] do you see the WST icon the same color of the copy besides it (should be deep navy 50% opacity)?
- [ ] on the post with a comment do you see the 1 comment in the same color of the WST copy (should be deep navy 50% opacity)?
- [ ] on the post without comments: do you see the icon and the copy besides it the same color as above (deep navy 50% opacity)?
- now hover on all those 3 elements
- [ ] all turns to deep navy 100% opacity?


bago2k4:
[Fix FoC hover footer icons](https://trello.com/c/JORTuPyb/2038-fix-foc-hover-footer-icons)

PR #879, opened 2019-05-21
Comment 😀 reactions

bago2k4:
Card: https://trello.com/c/6FWiIfWK

### Review with https://github.com/open-company/open-company-interaction/pull/33

To test:
- with user A add a comment and post it with the Send button
- [ ] with user A add another comment, use Enter while typing to add a new line
- [ ] now post it with CMD+Enter
- now add some more mixed comments from user A and B
- [ ] user A can edit/delete only his own comments?
- [ ] user B can edit/delete only his own comments?
- [ ] delete a comment work?
- edit comment is not working right now: medium-editor is instantiated but destroyed right after (save doesn't work because of it)
- share a comment button is in the code but commented out until we have comments addressability
- [ ] user A can react to one of his own comments from the picker?
- [ ] user B can react to one of his own comments from the picker?
- [ ] user A can react to one of his own comments using a reaction left from B
- [ ] user B can react to one of his own comments using a reaction left from A
- [ ] user A can remove a reaction from a comment?
- [ ] user A can react to the post with the picker?
- [ ] user B can react to the post with the same reaction left by B?
- check the feedback items left in the Trello card to see if they are all fixed
- test comments any other way you can think of
- test reactions any other way you can think of
- test comment reactions any other way you can think of


bago2k4:
[Comments - Part 1 (not yet threaded)](https://trello.com/c/6FWiIfWK/2022-comments-part-1-not-yet-threaded)

bago2k4:
@cjsauer good catch.
Ignore the edit comment for now, it's broken and i didn't have the time to debug it.
All the other issues should be resolved.
Please note that i forgot to open the Interaction PR, now that's in the description, please use that branch to test this.

bago2k4:
@cjsauer found the problem with the web client that was showing an error.
Also it's ok if a reactions shows a 0 sometimes, no need to change that for the moment.
Pull also Interaction Service branch since i fixed an issue there too.

cjsauer:
@bago2k4 one more thing: clicking emojis really fast actually results in unstable behavior and errors in the console. I was actually able to get the emoji count up to 3 with the same user. Maybe we should disable the button until the response is received. WYT?

PR #881, opened 2019-05-22
Comments addressability

bago2k4:
Card: https://trello.com/c/A0xsWskz

### Review after #879 

To test:
- add some comments
- copy the url of a comment
- open that URL
- [ ] do you see the post page open?
- [ ] do you see the page scrolling down to the comment?
- [ ] is the comment highlighted?
- [ ] does the bg color of the comment fade out after 5 seconds?
- navigate back to board
- open the post again clicking on FoC
- [ ] no comment highlighed this time?
- play with comments and their urls
- now add a new post with a user B mentioned in the body
- with user B click on the notification in bottom left corner
- [ ] post page was open?
- with user B leave a comment under the post
- with user A click the notification
- [ ] comment link was open and the page scrolled down highlighting the comment?


bago2k4:
@cjsauer sorry i pushed this in review too soon. Now it should be good though.

bago2k4:
[Link directly to a comment](https://trello.com/c/A0xsWskz/2048-link-directly-to-a-comment)

bago2k4:
@cjsauer yeah that's on purpose.
the problem is this: since the body is HTML it can contains media elements. For example a YT video, that can take a bit to load, so if the page scrolls down to the comment before the video is loaded (ie: as soon as the component is mounted) then when the video is loaded the comment will be scrolled down again. That's why i delayed so much the scroll+highlight.

I don't know if it's too much but since it's a limited thing (not everybody will land on a comment link) let's keep it on staging for a while and see if someone complains

We can merge this if everything checks out and maybe bring up this on tomorrow call.

belucid:
@bago2k4 just a thought, can't you detect the container change size on successful media load and adjust the scroll accordingly... such that you scroll early, then if media ends up loading later and changing the size, and therefore the scroll, you adjust it again?

bago2k4:
@cjsauer i merged in also this PR https://github.com/open-company/open-company-web/pull/886 to redirect to the comment highlight when a comment/interaction notification comes in.
So you can review this alone but it would be better to merge #886 first.
Also i have reduced the delay for the scroll+highlight from 5 seconds to 2.5.
Please take another quick look if you can.

bago2k4:
@belucid yeah i thought about a couple of solution for that delay, the main problem was the time when i brought up the discussion yesterday.
I'll take another look later today if i have time, for now i reduced the delay time (check comment above)

PR #887, opened 2019-05-24
Web font

bago2k4:
Card: https://trello.com/c/DlG0FUUt

We got the web font from Lineto foundry, i installed it but we need to make sure it's loaded and used correctly.
To check you can use WhatFont extension (should be available for all Chrome, FF and Safari) or check the computed style of the dev tools to see if it says "LL Circular Book Web" for book and "LL Circular Bold Web" for bold:
- open Chrome
- [x] check FoC fonts
- [x] check navigation sidebar fonts
- [x] check navbar fonts
- [x] check expanded post fonts
- [x] check menu panel fonts
- [x] check other panels fonts
- [x] repeat on FF
- [x] repeat on Safari
- [ ] repeat on Safari mobile (connect to desktop dev tools via cable)
- [ ] repeat on Chrome mobile (Android: you probably need to use you eyes here, not sure if there is something more accurate)


bago2k4:
[Circular web font](https://trello.com/c/DlG0FUUt/2063-circular-web-font)

PR #895, opened 2019-05-29
Settings data refresh

bago2k4:
Bug: we lost data refresh when opening a setting panel

To test:
- open org settings
- [x] do you see org data and team data refresh? Good
- open team management
- [x] do you see org data and team data refresh? Good
- open invite panel
- [x] do you see org data and team data refresh? Good
- open integrations
- [x] do you see org data and team data refresh? Good
- open user profile
- [x] do you see user data refresh? Good
- open user notifications panel
- [ ] do you see user data refresh? Good

### DIFF
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -592,8 +592,8 @@
     (api/get-auth-settings (fn [body]
       (when body
         (when-let [user-link (utils/link-for (:links body) "user" "GET")]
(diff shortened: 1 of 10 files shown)


cjsauer:
@bago2k4 it doesn't look like opening the Notifications panel is making a user request. Other than that all tests are passing.

bago2k4:
@cjsauer fixed, good catch, i added it to the wrong/old window (that will go away with cleanup)

PR #901, opened 2019-06-03
Org name max length

bago2k4:
Review with:
- [Storage](https://github.com/open-company/open-company-storage/pull/205)

To test:
- change `oc.web.utils.org:8` to be 60 instead of 50
- start web
- to go signup
- enter email and password
- when you get to `Tell us about you` step
- inspect the org name field
- change the max-length prop of the DOM input element to be 100
- now insert your first and last name
- insert a company name of 100 chars
- [x] can't do that? name is always cut at 60? Good
- now create the team
- [x] did you get an error about the org name? Good
- enter a company name of 50 chars
- [x] did it work? Good
- finish onboard
- go to settings
- change org name
- inspect the org name field and change the `max-length` property to be 100
- [x] can NOT change with an org name less than 3 chars?
- [x] can NOT change with an org name more than 50 chars?
- change again `oc.web.utils.org:8` to 50
- logout
- signup again
- [x] make sure the UI doesn't let you add a company name longer than 50 chars
- finish onboard
- go to settings
- [x] make sure the UI doesn't let you change the org name with something longer than 50 chars

Additional tests for routing library update (Secretary):
- go to /login (logged out, this ensure you are suing the React version of the app, not the static)
- dismiss the login popover
- nav to about from the top menu
- nav to pricing from the top menu
- signup via email
- signup via Google
- signup via Slack
- login via email
- login via Google
- login via Slack
- invite an email user and signup with that user
- reset password
- opening `/org/all-posts?org-settings=org` show the org settings panel?
- opening `/org/all-posts?user-settings=profile` show the user settings panel?
- opening `/org/all-posts?reminders=1` show the reminders panel?
- visit AP directly (no internal nav)
- visit a board directly (no internal nav)
- switch between boards, Drafts and AP
- share a post and copy the secure URL
- open it in incognito mode
- get a digest url (with id) and open it
- open a route that should 404


belucid:
@bago2k4 Long story, but I switched over to OpenJDK for Travis to work on other repos, it had trouble d/l'ing Oracle JDK anymore with its license changes. So as to not have Travis on a different VM than me locally, I switched my local VM to OpenJDK, but then the web app wasn't working. As part of trying to get it to work, I upgraded a few things... it turns out the key to get it working was to use specifically OpenJDK 8. In any case, I left the few things I upgraded as I went along in place (see the changeset now for this PR). The new cljs pointed out some fn private/public issues we had in the code.

Point of this note, other than to have you take a look at the current changeset as a 2nd pair of eyes, is if you run the web app now, there's a new warning that's from our code, not from dependencies:

`WARNING: Use of undeclared Var oc.web.actions.nav-sidebar/show-reminders at line 86 src/oc/web/utils/user.cljs`

Any idea how to fix that? We seem to be using that fn by full namespace vs. a requires. Was that for a reason? To avoid a circular dependency or something?

belucid:
@bago2k4 not sure if this test step is right, but using `?org-settings=main` on the URL does nothing.

bago2k4:
the `?org-settings=main` test step was wrong, it should be `?org-settings=org`

PR #905, opened 2019-06-04
Fix FoC timestamp tooltip

bago2k4:
We lost the edit info in the timestamp tooltip.

Part of this card: https://trello.com/c/CJahkpWQ

To test:
- go to AP
- create a new post and publish
- [x] tooltip has only publish date/time and no user (user is implicit from the FoC)
- edit the post
- [x] tooltip still the same (edit are shown only if they are 24 hours later of publish)
- edit a post published more than 24 hours ago
- [x] tooltip shows publish and last edit date with user of edit if different?
- expand a post
- [x] timestamp still the same?


bago2k4:
[Sort and filter posts](https://trello.com/c/CJahkpWQ/1992-post-sorting)

belucid:
@bago2k4 I don't get any FoC tooltip an author.

bago2k4:
@belucid added

belucid:
@bago2k4 with a new post I see a Posted and Edited. Not just a Posted.

https://take.ms/hoRJbV

bago2k4:
@belucid what you mean with a new post? can you do a quick loom or at least a screenshot of the tooltip?
I tried now and until i publish the post the tooltip in drafts and then in AP always says "Posted on..." no edit. Then if i edit still only that (need to edit at least after 24h)

belucid:
@bago2k4 we need to show the edit, even if < 24h, if the author and editor aren't the same.

bago2k4:
@belucid fixed

PR #908, opened 2019-06-05
Preload interstitial images

bago2k4:
Part of: https://trello.com/c/CJahkpWQ

NB: probably better to test this on staging to make sure the load time is congruent with production

To test:
- load AP on desktop
- do you see the interstitial image before the posts show up?
- test the same on mobile


bago2k4:
[Sort and filter posts](https://trello.com/c/CJahkpWQ/1992-sort-and-filter-posts)

PR #910, opened 2019-06-06
Board pagination [WIP]

bago2k4:
Part of: https://trello.com/c/CJahkpWQ

Review with:
- [Lib](https://github.com/open-company/open-company-lib/pull/66)
- [Storage](https://github.com/open-company/open-company-storage/pull/207)

Possible scenarios to fix and then test:
- direct load of a post within the first page
- direct load of a post out of the first page
- internal nav to a post within the first page
- internal nav to a post out of the first page (with a notification click for example)
- make sure pagination is not trying load the last page over and over but only one time

PR #911, opened 2019-06-10
Post sorting

bago2k4:
Review after #910 

Part of https://trello.com/c/CJahkpWQ

Review with:
- [Storage](https://github.com/open-company/open-company-storage/pull/208)
- [Change](https://github.com/open-company/open-company-change/pull/21)
- [Lib](https://github.com/open-company/open-company-lib/pull/66)
- [Lib 2](https://github.com/open-company/open-company-lib/pull/67)
- [Infra](https://github.com/belucid/open-company-infrastructure/pull/109)

To test:
- boards pagination
- boards sort
- AP pagination
- AP sort
- NEW comment(s) tag
- drafts board (no sort, show always all the entries, no attempt to load more)
- sort dropdown style
- sort dropdown on mobile
- check 404 for not existing post uuid
- check current sort type is remembered and reused across sections and AP
- check live comments coming in from another user
- check section name update doesn't break pagination or section data: try switching sort after a board edit
- check pagination doesn't break when switching sort order: load more pages then switch sort then switch back
- check post move still work as expected
- check post delete still work as expected

Possible scenarios inherited from board pagination PR to fix and then test:
- direct load of a post within the first page
- direct load of a post out of the first page
- internal nav to a post within the first page
- internal nav to a post out of the first page (with a notification click for example)
- make sure pagination is not trying load the last page over and over but only one time


bago2k4:
[Sort and filter posts](https://trello.com/c/CJahkpWQ/1992-sort-and-filter-posts)

bago2k4:
@belucid additional test steps for the public post link issue discovered by Stuart:
- copy the private share url of a post
- open it with a user of that team and make sure you see it
- open it with a user of another team and make sure don't see it
- open it when in incognito and make sure you get the login wall
- copy the public share url of a post
- open it with a user of that team and make sure you are redirect to the real post url
- open it with a user of another team and make sure don't see it
- open it when in incognito and make sure you get the read only post view

belucid:
@bago2k4 conflicts here

bago2k4:
@belucid no anymore

belucid:
@bago2k4 pagination is working fine on a board, but not on AP. On AP it never paginates.

The networking calls all happen and look good. But the UI never does anything with the additional posts.

belucid:
@bago2k4 bug sequence:

- Have posts w/ a natural sequence by created-at, and adjust the sequence with a comment from another user so you can tell which sort is operative
- Go to all posts w/ a user that didn't make the comment
- Select "Recently posted" as your sort order
- Expand a post
- Click back to AP
- Notice that even though the selection stayed at "Recently posted" your posts are now in Recent activity order

belucid:
@bago2k4 the comment count increments live, but comment NEW badge doesn't come in live correctly. I actually see it flash briefly then go away.

Sequence:

- 2 users, user A is looking at AP with a post in view, user B has the the post expanded
- Leave a comment as user B
- Notice the comment count increments for user A, notice (NEW) is there for a brief moment then disappears
- Refresh AP for user A, notice the (NEW) is back

src/oc/web/actions/activity.cljs
```diff
@@ -955,3 +984,7 @@
                                                :dismiss true
                                                :expire 3
                                                :id (if success :mark-unread-success :mark-unread-error)})))))
+
+(defn change-sort-type [type]
+  (cook/set-cookie! (router/last-sort-cookie (router/current-org-slug)) (name type) (* 60 60 24 6))
```
belucid: is that 6 days? Why so short?

belucid: @bago2k4

src/oc/web/actions/org.cljs
```diff
@@ -67,6 +67,23 @@
       (cook/remove-cookie! (router/last-org-cookie)))
     (routing-actions/maybe-404)))
 
+(def default-board "all-posts")
+
+(defn get-default-board [org-data]
+  (let [last-board-slug default-board]
+    ; Replace default-board with the following to go back to the last visited board
+    ; (or (cook/get-cookie (router/last-board-cookie (:slug org-data))) default-board)]
```
belucid: @bago2k4 commented out?

src/oc/web/actions/activity.cljs
bago2k4: Good catch, i copied the code we have been using for all the other cookies: last used org (if you have multiple, last used board, last board etc..
We can definitely use something longer here. Any suggestion?

src/oc/web/actions/org.cljs
bago2k4: Yeah that's because we want to bring the user always back to AP, we used to bring the user back to the last visited board or AP if nothing was specified in the url.
Want me to remove this code? I think it can still come in handy

belucid:
How about 12 months.

belucid:
Thanks. It's fine to keep.

bago2k4:
@belucid fixes are in:
- NEW tag disappearing is fixed: as i anticipated yesterday night the problem was the entry API that wasn't returning the :new-at value (last timestamp created-at) so when a new comment was coming in the NEW appeared but after an entry reload it would disappear
- default expiration time changed to 365 days for all optional/preferences cookies: i left 60 days for the JWT cookie since it has always worked and i don't see why we should change it

I was wondering if the `:new-at` key is a bit too weird/cryptic as name, do you think i should change it with something more specific like: `:last-comment-at` (even if it filters out the comments of the current user)?

belucid:
@bago2k4 I sort of like the ambiguity of `new-at` because it suggests the rule for "newness" has been applied, and here's the date of it. We can change the rule, such as to include reactions for example, or to not exclude the users own comments, or whatever, and the name still works. Whereas if we use something like `last-comment-at` and then we change our mind later as to what makes a post new for the current user, we are left with a name that no longer works.

belucid:
@bago2k4 private share URLs aren't working for logged in users

belucid:
@bago2k4 issue with Draft badging

sequence:

- start w/ no drafts
- go to a section
- create a post for that section
- allow the post to auto-save, notice Draft count goes to 1
- now post the post
- notice draft count stays at 1
- navigate to another section, navigate to AP, draft count still 1
- navigate to drafts, or refresh and notice draft count is cleared as it should be

PR #912, opened 2019-06-12
Fix login redirect

bago2k4:
BUG: if you try to login with the wrong password you get redirected to the homepage. This is because the failed login return a 401 and we force the user to logout on every 401 response from our services since that means the token is no longer valid.

Solution: i added a check so the force logout and redirect happens only if the user is actually logged in, this should keep the old behaviour but avoid redirect when the user tries to login

To test:
- signup with email
- finish onboard
- logout
- try logging in with a wrong password
- [x] you still on the login form?
- [x] do you see a failed login error?
- now try to login with the real password
- [x] did it work?
- now with dev tools go and change the `jwt` cookie (it's called `localhost-jwt` or `staging-jwt` or `jwt` depending on the env)
- replace it with a random string
- refresh the page
- [x] did it NOT load?


belucid:
So @bago2k4 ...  per our Slack discussion, to verify we didn't regress at all here, I created an org w/ a public section.

I then got a normal link to a normal post and tried it use the link w/o a login at all. On mainline I end up at the login page (with the URL still of the post). Only real problem is I see the "Internal error occurred"  error dialog. That's not good.

On this branch, the same thing happens. So, not an issue. I put the network dialog error in the feedback card for later.

belucid:
@bago2k4 if you put nothing into the form, and click "Sign in" you get a failed login attempt and "System troubles logging in. Please try again, then contact support.". Login just shouldn't be attempted if email or password are blank. Sign In button should be disabled.

belucid:
@bago2k4 this happens at both the login overlay and the login wall (where the button is "Continue").

bago2k4:
@belucid now the login button are disabled in both login wall (`activity-not-found` component) and in the login overlay (`oc.web.components.ui.login-overlay/login-with-email`).
The error you saw is due to us trying to load the comments.. i should fix that but not very connected to this..
let me know if you want me to fix it now or add a new card

PR #915, opened 2019-06-13
Fast copy changes 2

bago2k4:
Card: https://trello.com/c/Svdv1puR

Items:
- Reduce the highlighted space to 16px (looks a lot larger): https://www.dropbox.com/s/w61c1t4ohpzl3qo/Screenshot%202019-05-10%2010.18.44.png?dl=0
- Remove background color on links found within post comments
- Pagination text tweak: Remove italics, change to deep navy @ 50%
- Change line-ight on the full-screen post headline to 40px (desktop only)
- Make link interactions consistent between post links and comment links: https://www.dropbox.com/s/eqgun7surbfvd4x/Screenshot%202019-06-04%2015.22.56.png?dl=0
- The "move post" menu dissappears if hover off the FoC of the post you're trying to move (should stay open always once the menu is opened when trying to move a post)
- When in all posts, and I move a post, it vanishes from the feed
- Change line-height on the post title (desktop, full-screen editing experience only) from 40px to 34px (https://www.dropbox.com/s/umz7v6x369s3bzn/Screenshot%202019-06-10%2013.10.24.png?dl=0)
- Use a regular link to a post while you're not logged in at all, you get a login page (that's good!), but you also see the "Internal error occurred" error dialog. That's not good.
- Create section allows there to be a newline in the name, this then breaks stuff downstream (settings tooltip, bolding in Slack messages, etc.). To see this, create a new section, type the section name, then press enter when you're done typing the name.
- Mobile: WST header bar should be sticky when the modal is open so that you can close the window without having to scroll all the way back to the top https://www.dropbox.com/s/zfe7j6ggxvzwxos/Screenshot%202019-06-12%2008.48.34.png?dl=0
- Error dialog from trying to load comments when not logged in. Scenario: normal link to a normal post and try to use the link w/o a login at all. End up at the login page (with the URL still of the post). Problem is you see the "Internal error occurred" error dialog.
- When a user "sends" a post from the WST panel, please hold the message "Sent to...X" until the user leaves the panel and returns (right now it's easy to lose what members have been sent a reminder vs who hasn't)
- Create a new reminder with the reminder service unavailable and you get the network error, but it creates the reminder in app state anyway and closes the form, better to not "create it" and leave the form open.
- Change the FoC headline to (desktop stream view only): font-size: 24px; line-height: 28px;
- I bolded a link and this happened: https://www.dropbox.com/s/ahidjagsbpn35dl/Screenshot%202019-06-20%2014.00.15.png?dl=0


bago2k4:
[Fast Tweaks](https://trello.com/c/Svdv1puR/2000-fast-tweaks)

belucid:
@bago2k4 kibit gripe on this one.

bago2k4:
@belucid fixed

belucid:
@bago2k4 can no longer type normally into the section name text field

sequence:

- open to the section name text field (via edit or add)
- type at a normal (quick) speed
- notice that many characters are dropped out of what you type and you end up with gibberish

belucid:
@bago2k4 the rest looks good other than the issue above

bago2k4:
@belucid fixed

belucid:
@bago2k4 2 more issues related to section name changes. After you change the name the tooltip still shows the old name. Also we get both our tooltip and the browser's tooltip:

https://take.ms/lio3w

bago2k4:
@belucid good catch!

the bootstrap tooltips are a bit of a hassle for this. They don't update themselves but also they let the browser show the default tooltip if you change it but don't update it via the js call.

Not sure if there is something better around: as simple as them, work on all browsers and versions but w/o this little issue.

PR #920, opened 2019-06-17
Fix body truncation

bago2k4:
From: https://trello.com/c/lu985GuX
item:
> Make sure abstract doesn't get truncated in FoC, only body (maybe increase the body truncation to match the max possible length of the abstract?)

To test:
- add a post with a long abstract composed by very long word (like a word long one line: need this to make sure the abstract will take more than 4 lines)
- is it truncated when shown in FoC?
- now add a post with long body but no abstrasct
- is the body used in FoC?
- is it truncated to 3 lines?
- does it has a ... at the end?
- now add a post with no body and abstract
- do you see the post in FoC?
- no errors?


belucid:
nice!

PR #922, opened 2019-06-18
Web client cleanup

bago2k4:
Card: https://trello.com/c/ZKKgEHi2

### Review after #920 

NB: the following items:
> Remove comments preload from `stream-item` component
> Remove `check-item-read` function from `stream-item` on mount
Where done in this PR #920 

For the most part testing this means switch to this branch, stop the client boot task and re-build from scratch. Check that no warnings are shown (except one from `cuerdas` and one about `oc.web.actinos.nav-sidebar/show-reminders`).

To test:
- go to ziggeo and add a callback to the default app to your local storage service
- build web mainline
- add a post with a ziggeo video
- no stop web service
- switch to this branch
- re-build the web project
- open the web client
- do you see the ziggeo video preview?
- can you play it when in the expanded post view?
- take a test ride of the app: login, signup, AP, section, edit section, settings, user settings, expanded post, reminders, public post link and whatever you can think of


bago2k4:
@cjsauer good catch! should be fixed now, i also found another place where the `moment-timezone` package was used but not included..
should be good now

PR #933, opened 2019-06-27
Improve initial load

bago2k4:
#### Review after #927 

Try to delay the loading of all the resources that are not immediately needed.
It delays by 2.5 seconds:
- all the sections both sorts except the one currently selected
- AP both sorts if it's not currently selected
- team, roster and channels: delay only if the app is not loaded with the one of the settings panel open, also roster and channels call where particularly expensive for Slack because they include a sync call to Slack apis

To test:
- open staging AP (with dev tools closed)
- now refresh and time how long it take to show the posts
- same for a section (the one with more posts)
- now deploy this branch
- repeat the timing for the 3 urls above
- [x] did it improve?
- [x] now after loading AP or a section make sure the settings panels are all working
- [x] now load `/carrot/all-posts?org-settings=org` and make sure the settings panel is loaded as expected
- [x] now load `/carrot/all-posts?org-settings=integrations` and make sure the settings panel is loaded as expected
- [x] now load `/carrot/all-posts?org-settings=team` and make sure the settings panel is loaded as expected
- [x] now load `/carrot/all-posts?org-settings=invite` and make sure the settings panel is loaded as expected
- [x] now load `/carrot/all-posts?user-settings=profile` and make sure the settings panel is loaded as expected
- [x] now load `/carrot/all-posts?user-settings=notifications` and make sure the settings panel is loaded as expected
- [x] now load `/carrot/all-posts?reminders=true` and make sure the settings panel is loaded as expected


bago2k4:
@belucid the issue you found with mention users on a post page was actually on this branch.
As i explained the list of users is not immediately ready and it was failing to setup the ME plugin.
Now should be fixed but please test the mentions in all these places:
- add comment (go to post page with internal nav and direct load)
- edit a comment (same as above)
- add a post (from a section, from AP and from a post page)

belucid:
@bago2k4 I still don't get mentions working in comments loading the browser directly at a post.

- Works: Start from AP, expand post, comment
- Works: Start from section, expand post, comment
- Doesn't work: Start from post (e.g. `http://localhost:3559/carrot/decisions/post/c95d-4cdf-aa02`), comment

bago2k4:
@belucid fixed the issue above, it was a `when` with wrong closed bracket.
Also adding these test steps:
- load the app with this url `/carrot/all-posts?new=1`
- try to use the mentions in the body as soon as possible
- did it work? Good

PR #935, opened 2019-06-27
Comment buttons tooltips

bago2k4:
Card: https://trello.com/c/DCsAsrtG

This improve the handling of all bootstrap tooltips in the app using a new mixin that make sure the tooltips are initialized when a component is mounted and when the props are updated they are re-initialized (to make sure the new titles are picked up).

This started as an issue on the comment buttons but i finally end up applying the same thing everywhere the tooltips are used.

To test:
- go to a post page
- add a comment
- immediately hover the reaction button (you get only one button immediately after posting)
- [x] does it say Add reaction?
- wait for the other buttons to appear
- [x] does every button have the right tooltip?
- [x] check the reactions's tooltips (before and after leaving reactions)
- [x] check the tooltip in the more menu (FoC horizontal ellipses button)
- nav to a section
- check the section COG tooltip
- click it
- change section name
- save
- [x] check the section COG tooltip again
- check every other tooltip you can think of


bago2k4:
[Bootstrap tooltips fixes](https://trello.com/c/DCsAsrtG/2104-bootstrap-tooltips-fixes)

PR #938, opened 2019-07-03
Upgrade sentry

bago2k4:
The Sentry SDK has had substantial updates lately. They left behind the Raven SDK and switched to a new Sentry SDK that is used in a similar way but needs some changes to the code to be adopted.

I thought we should keep up with it to avoid losing error reports so i tried to update it.

To test:
- run locally
- make sure you have a web DSN configured in local-settings
- from console run `oc.web.lib.sentry.capture_message_BANG_("This is a test message")`
- go to sentry.io
- [ ] do you see the error? Good
- make sure the app still works (test main functionality)
- optionally you can insert a runtime error in the code and make sure it's reported to sentry


belucid:
@cjsauer can we get this review done?

PR #956, opened 2019-07-17
App update banner refresh

bago2k4:
Card: https://trello.com/c/z62uhpLg

To test:
- change line 10 and 11 of `oc.web.actions.app-update` and set 3 and 60 seconds for example
- open the app
- can you see the app update happening every 3 seconds? Good
- now delete this file `rm resources/public/version/version.json` and wait for rebuild
- open the web app
- do you see the banner to refresh the app? Good
- dismiss it with the X
- do you NOT see the app update refresh every 3 seconds anymore? (it should be delayed by 60 seconds now)
- [ ] after 60 seconds do you see the banner again? Good
- [ ] try the refresh button


src/oc/web/actions/app_update.cljs
```diff
@@ -0,0 +1,62 @@
...
+  "App update notification was dismissed, let's set a new timeout with a longer wait time."
+  []
+  ;; Clear the old interval
+  (when @app-update-timeout
+    (.clearTimeout js/window @app-update-timeout))
+  ;; Set a new interval
+  (set-app-update-timeout default-app-update-dismiss-interval))
+
+(def update-verbage
+  (if js/window.isDesktop
```
cjsauer: @bago2k4 as of #972 this will no longer be referred to as `js/window.isDesktop`, but we should instead be checking for the existence of the `js/window.OCCarrotDesktop` object. #972 contains changes that scopes Electron's JS injections under this object in order to avoid any future collisions on the `window` object.

cjsauer: We can likely add this PR as a dependency for #972, which would allow us to make the fix in #972 instead of in this branch.

src/oc/web/actions/app_update.cljs
```diff
@@ -0,0 +1,62 @@
+(ns oc.web.actions.app-update
+  (:require [oc.web.api :as api]
+            [oc.web.actions.notifications :as notification-actions]))
+
+(defonce app-update-cicle (atom false))
```
cjsauer: @bago2k4 "cicle" is actually spelled "cycle" with a `y`

bago2k4:
@cjsauer fixed the typo and also renamed the namespace using the `web` prefix to avoid confusion with the electron app-update namespace.

Finally i decided to avoid using a single update namespace for both since it was requiring more time than i thought to unify those: we are both using these namespaces as singletons.

cjsauer:
@bago2k4 
>I think we should add some timbre logging to the app-update namespace so we can see when app updates are occurring (or failing to occur).
WYT?

bago2k4:
@cjsauer added logs

PR #964, opened 2019-07-24
Default to current section

bago2k4:
Card: https://trello.com/c/lQolzRKk

Bug: if you nav to a section and then press the New post button the pre-set section is the last used section or the first in alpha order. It should be the currently selected section instead. It already works this way if you start from quick post.

To test:
- let's say your org has: Section 1, Section 2, Section 3 in this alpha order
- go to AP
- click New post
- change section to Section 2
- add a post
- publish it
- now nav Section 3
- click New post
- [x] default section is Section 3?
- go to Section 3
- now add a Section 4
- nav to it
- click on the Create post button in the empty section placeholder
- [x] default section is Section 4?
- now go to Section 2
- click to expand quick post
- [x] default section is Section 2?
- now add a private section (Section 5) with another user
- add the current user as viewer only
- with the current user nav to the private section
- expand quick post or click New post
- [x] default section is still Section 2 (or whatever the last post was published in but not Section 5)?


bago2k4:
[New post default section](https://trello.com/c/lQolzRKk/2149-new-post-default-section)

PR #965, opened 2019-07-24
Fix board watch

bago2k4:
Card: https://trello.com/c/emcPfpRb

Bug: currently the watch request for CS is sent every time a section or AP is loaded, it means that the user is watching always the last loaded thing. Also the read data were loaded indiscriminately, that could lead to the NEW comments badge problems, but i wasn't able to replicate that.

This also adds some debug info to the FoC node needed in case the NEW thing happen again.

To test:
- with User 1 go to AP
- with User 2 add a post to a section  User 1 has access to
- [x] User 1 gets the new post? Good
- with User 2 edit the post and change headline
- [x] User 1 gets the new headline? Good
- User 1 nav to Section 1
- User 2 add a post to Section 2
- [x] nothing happen in User 1 session
- User 2 update the post in Section 2
- [x] nothing happen in User 1 session
- User 2 add a post in Section 1
- [x] User 1 sees the new post coming in live
- User 2 updates the headline of the post in Section 1
- [x] User 1 sees the new headline


bago2k4:
[New comments badge disappear and then re-appear](https://trello.com/c/emcPfpRb/2147-new-comments-badge-disappear-and-then-re-appear)

PR #975, opened 2019-08-01
Fix abstract autoresize field

bago2k4:
Bug: abstract field is broken on production right now since it doesn't autoreszie.

To test:
- add a post in fullscreen mode
- [x] does abstract field autoresize?
- publish it
- now start a new post from quick post
- switch to fullscreen
- [x] abstract field is empty?
- type something in the abstract field
- [ ] abstract autoresizes?
- publish post again
- click New post
- [x] abstract is empty again?

### DIFF
--- a/src/oc/web/components/cmail.cljs
+++ b/src/oc/web/components/cmail.cljs
@@ -630,7 +630,6 @@
               [:textarea.cmail-content-abstract.emoji-autocomplete.emojiable.group.oc-mentions.oc-mentions-hover
                 {:class utils/hide-class
                  :ref "abstract"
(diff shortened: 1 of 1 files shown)


cjsauer:
@bago2k4 test cases are passing, but I'm noticing that the green `Post` button always shows the 🚫 symbol on hover, even when I've filled in every field. This doesn't appear to happen on `mainline`, so I think this change might be the culprit.

cjsauer:
@bago2k4 never mind, I think I have to focus a different field for that button to update its state. Likely unrelated to this change.

PR #983, opened 2019-08-16
Replace static-js useragent detection with GCL implementation

cjsauer:
Relevant Trello card:  https://trello.com/c/9N5ncjab/2161-mobile-web-app-redirect

This PR doesn't completely cover the card linked above, but rather supports its development directly.

## Testing

First, have a look at the [new useragent namespace](https://github.com/open-company/open-company-web/blob/2b43b3db4c196a1d965dbff0447ea2c5bd769284/src/oc/shared/useragent.cljs) and get a sense for the platform/browser/OS detection functions.

Now, on as many combinations as you have available, use `oc.shared.useragent.*` from the JS console and verify that the results match what is expected. For example, open Firefox on Mac and observe:

- [x] `oc.shared.useragent.mac?` returns true?
- [x] `oc.shared.useragent.firefox?` returns true?

etc etc


cjsauer:
[Cross-Platform Mobile app (iOS / Android) MVP](https://trello.com/c/bvFHsWWZ/2060-cross-platform-mobile-app-ios-android-mvp)

cjsauer:
[Mobile web app redirect](https://trello.com/c/9N5ncjab/2161-mobile-web-app-redirect)

belucid:
Is this ready for review @cjsauer ?

cjsauer:
Just needs some test steps.

cjsauer:
@belucid should be good now.

src/oc/web/actions/jwt.cljs
```diff
@@ -15,7 +16,7 @@
 
 (defn logout
   ([]
-   (logout (if js/window.OCCarrotDesktop
+   (logout (if ua/pseudo-native?
```
bago2k4: @cjsauer you sure this redirect should be changed for both desktop and mobile wrappers? Before it was only for desktop if i understand the removed code...

src/oc/web/components/ui/onboard_wrapper.cljs
```diff
@@ -145,7 +146,7 @@
             "Sign up"]]
         [:div.footer-link
           "Already have an account?"
-          [:a {:href (if js/window.OCCarrotDesktop
+          [:a {:href (if ua/pseudo-native?
```
bago2k4: Same here

src/oc/web/core.cljs
```diff
@@ -534,7 +535,7 @@
       (timbre/info "Routing logout-route" urls/logout)
       (cook/remove-cookie! :jwt)
       (cook/remove-cookie! :show-login-overlay)
-      (router/redirect! (if js/window.OCCarrotDesktop
+      (router/redirect! (if ua/desktop-app?
```
bago2k4: If the other are correct then this is wrong?

src/oc/web/actions/jwt.cljs
cjsauer: @bago2k4 yes it's correct. Both desktop and mobile need to redirect to the "desktop login" page, which is really just a plain login page. The name `desktop-login` is kind of unfortunate. I think we should rename it here.

src/oc/web/core.cljs
cjsauer: @bago2k4 good catch. This one is incorrect.

PR #987, opened 2019-08-21
Fix org save error

bago2k4:
Bug: login as admin, open org settings panel, change org name, hit save: :boom: error notification but the operation actually worked.

To test:
- verify the above situation is failing on mainline
- switch to this PR
- repeat the steps above
- [x] worked? Good
- now update org avatar
- add email domain
- remove email domain
- [x] everything fine?
- notifications are always inline with what is happening?


belucid:
@bago2k4 I don't get any notification upon successfully adding an email domain (I do on remove).

PR #990, opened 2019-08-29
Desktop tweaks

cjsauer:
Trello: https://trello.com/c/pPG7lrg7/2181-desktop-tweaks

## To test

- [x] Is the desktop icon the standard Carrot logo we all know and love?
- [x] Is the window width/height constrained in such a way to prevent responsive issues?
- Login with user A on the desktop app, and user B in your browser
- Try mentioning user A in a comment by user B. Because user B is using the browser, the desktop app cannot be focused.
- [x] Did you receive a growl notification?
- Now, focus the desktop application window, and then trigger another growl notification. I did this by using `setTimeout(function() { temp1.click() }, 5000)` in the browser console, where `temp1` is the button element for "Add Comment" or "Add Reminder". The main point is that you want to be able to focus the desktop application while triggering an event that would normally display a growl notification.
- [x] Did you **NOT** receive a growl notification? Good!


cjsauer:
[Desktop Tweaks](https://trello.com/c/pPG7lrg7/2181-desktop-tweaks)

resources/electron/renderer.js
```diff
@@ -23,3 +23,8 @@ window.OCCarrotDesktop.setBadgeCount = function(count) {
   console.log("Sending set-badge-count IPC: " + count);
   ipcRenderer.send('set-badge-count', count);
 };
+
+window.OCCarrotDesktop.windowHasFocus = function() {
+  console.log("Determining focus state of desktop window");
```
bago2k4: Did you leave this in on purpose?

src/oc/electron/main.cljs
```diff
@@ -54,6 +54,8 @@
                       win-frame-settings
                       {:width w
                        :height h
+                       :minWidth 1280
```
bago2k4: Isn't a bit too much 1280 as minimum width? If you have a mac 13 inch it should be good but what about all the smaller laptop?
I think we should set it to 1024 for example, WYT?

src/oc/electron/main.cljs
```diff
@@ -158,4 +160,7 @@
   (.on ipc-main "set-badge-count" (fn [event arg] (.setBadgeCount app arg)))
   (.on ipc-main "show-desktop-window" (fn [event arg] (when @main-window
                                                         (.show @main-window))))
+  (.on ipc-main "window-has-focus?" (fn [event]
+                                      (let [ret-value (boolean (.getFocusedWindow BrowserWindow))]
+                                        (set! (.-returnValue event) ret-value))))
```
bago2k4: Let me see if i get this.
When you call `ipcRenderer.sendSync('window-has-focus?')` it generates an event.
Then here you capture that event and add a boolean value to it that says if the window is focused, that correct?

bago2k4:
Also @cjsauer for the tests, where can i download a version of Carrot desktop with these changes? Is it the latest SNAPSHOT release?

bago2k4:
Hmmm i tried 1.1.5-SNAPSHOT and 1.1.6 and none has the resize restriction: not on the external frame, not on the internal window view.

resources/electron/renderer.js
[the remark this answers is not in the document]
cjsauer: @bago2k4 I keep these logs in each renderer IPC call to make it easier to trace what's happening between renderer and main process. It's nice to be able to know when IPC calls are happening in the console.

src/oc/electron/main.cljs
cjsauer: @bago2k4 originally I did set it to 1024, because that felt like a natural number, but the responsive view didn't look quite right at that size. It created a horizontal scroll bar. I can change it back for you to see it.

src/oc/electron/main.cljs
cjsauer: @bago2k4 Yeah, the renderer process generates a synchronous event (blocks the UI), and the main process catches that event above, and sets the `returnValue` field on the event to the focus state of the window. The `returnValue` field is how Electron responds to synchronous events.

cjsauer:
@bago2k4 ah I forgot to release a SNAPSHOT version for you. I'll do that right away.

cjsauer:
@bago2k4 
Here is the draft build for this PR: https://github.com/open-company/open-company-web/releases/tag/untagged-bde7065c04fe58ff095a

PR #992, opened 2019-08-30
Fix notifications expiration time

bago2k4:
Card: https://trello.com/c/uDDHbrdu

To test:
- check the code changes and think if the nature of the notification is correct with expiration time associated
- you can try a few notifications to see if the expire time is respected


bago2k4:
[Notification timing](https://trello.com/c/uDDHbrdu/2166-notification-timing)

PR #997, opened 2019-09-02
FoC icons

bago2k4:
Card: https://trello.com/c/OQ23Jg1H

To test:
- add a new post with an attachment
- publish it
- [x] do you NOT see an icon besides the number of viewers in FoC?
- [x] do you NOT see an icon besides the number of attachments in FoC?


bago2k4:
[FoC WST and attachment icons](https://trello.com/c/OQ23Jg1H/2206-foc-wst-and-attachment-icons)

PR #1000, opened 2019-09-02
DnD images into comments

bago2k4:
To fix: there is no indication that a file is uploading, we need to address this in a later card since it's going to take a bit more.

To test:
- focus on a comment field
- drag in an image
- [ ] did it work?
- now drag in another file that is NOT an image
- [ ] did it NOT work?


bago2k4:
Closing this and adding test steps to #1008 that is branched from this.

PR #1002, opened 2019-09-02
Performance improvement

bago2k4:
Card: https://trello.com/c/vJY8xS8Z

Review with
- [Storage](https://github.com/open-company/open-company-storage/pull/217)
- [Interaction](https://github.com/open-company/open-company-interaction/pull/37)
- [Lib](https://github.com/open-company/open-company-lib/pull/72)

To test:
- open AP
- as soon as you see it click on the first post
- [x] do you see the ghost posts but not the shaking carrot again?
- do you see 7 posts loaded per time?
- now use responsive mode from Chrome dev tools and make your window very tall that it fits at least the first 7 posts
- refresh the page
- [x] do you see the app immediately loading the second batch of posts?
- [x] can you scroll down and load more?
- add a reaction (should be immediate, not waiting for the network request
- add a comment (should be immediate, not waiting for the network request
- add a reaction to a comment (should be immediate, not waiting for the network request
- now open a post url
- add a comment
- [x] do you see the new comment being added to the list right away and never disappearing from the list?
- go to Network tab of Chrome dev tools and change the dropdown to 3G Slow
- add a comment
- [x] do you see the new comment being added to the list and never disappearing (neither for a small amount of time)?
- ngo to Network tab of Chrome dev tools and change the dropdown to Offline
- add a comment
- [x] do you see the comment going immediately back into the edit field?
- [x] repeat with normal network and edit a comment (instead of add)
- [x] repeat with 3G Slow and edit a comment (instead of add)
- [x] repeat with Offline and edit a comment (instead of add)
- remove the Network filter from dev tools
- force show the new version notification (in `oc.web.actions.web-app-update` remove 1000 from line 12 and 13, also change line 30 to `(if (not= status 404)`)
- load the app
- wait for the new version notification
- click it
- [x] do you see the shaky carrot appear while it waits for the page refresh?
- [x] does the click action seems more immediate than it used to be?
- now open a post with some comments
- turn off your network
- add a comment
- [x] do you see the comment being reinserted in the comment field?
- [x] do you still see the comments above the add comment field?

Multiple sessions test (test for interaction/lib changes):
- open 2 tabs on the same browser windows on the same post url
- now split the tabs in 2 windows (start from the same windows to make sure they are sharing the same cookies)
- add a comment from one of the 2
- [x] do you see the comment being added to the other window in real time?

Regression test:
- open AP
- click in quick post
- add some body
- refresh page
- [x] does it open back into edit mode with your changes visible?
- add some more text and some abstract and headline
- now refresh again from full post view
- [x] does it open back into full post view with the latest changes?


cjsauer:
@bago2k4 
>do you see the app immediately loading the second batch of posts?
This isn't happening...I see `Loading more posts...` at the bottom, and then the next 7 come in after a small delay.

cjsauer:
@bago2k4 
>do you see the new comment being added to the list and never disappearing (neither for a small amount of time)?

This also isn't the case for me. When using Slow 3G the comment disappears for a long time, and then finally comes back.

bago2k4:
> @bago2k4
> 
> > do you see the app immediately loading the second batch of posts?
> 
> This isn't happening...I see `Loading more posts...` at the bottom, and then the next 7 come in after a small delay.

Yeah that's what's meant to happen, we need to render the first 7 posts to see if we need to load more so you see the first 7 appear then loading more and then the next batch is shown.
No need to scroll down to get the rest of the posts.

PR #1011, opened 2019-09-05
File upload size

bago2k4:
Make sure we are using the same size for files uploaded via FileStack picker and via DnD.

Review with: https://github.com/belucid/open-company-infrastructure/pull/116

To test:
- new post
- click image button
- pick an image that is > 5MB
- [x] did it work?
- now try with a file that is > 20MB
- [x] did it NOT work?
- now do the same via DnD
- now do the same via DnD in a comment


cjsauer:
@bago2k4 DnD file upload of valid files appears to get stuck at 0%...files over the limit are correctly denied tho.

PR #986, opened 2019-08-21
Sentry upgrade and catch wrapper

bago2k4:
NB: wait for this cljsjs PR to be merged in https://github.com/cljsjs/packages/pull/1941

Code to add:
`(throw (js/Error. "An error occurred..."))`

To test:
- make sure you have a Sentry DSN configured for the web project where you plan to test this
- in a component you are sure that is being mounted add the above code line in the `did-mount` handler
- load the app
- :boom: you should see an error (or more) in the console with the above message
- check on sentry.io that the error is reported
- merge

PR #1017, opened 2019-09-09
Fix fullstory session urls.

cjsauer:
**_(Re-open of #999, accidentally merged before having reviewed #986. `mainline` has been correctly reverted.)_**

Review after #986 

No need to test this, we will see when sentry comes in if it works, right now they don't have the fullstory session URL. Just check the code is right and no warnings or errors appear when building this branch


belucid:
conflicts here @cjsauer and @bago2k4 

Where are we at with this one?

bago2k4:
@belucid it was blocked because waiting for #986 that has been merged yesterday, now it's good to go.

PR #1023, opened 2019-09-11
Fix follow-up copy.

bago2k4:
Card: https://trello.com/c/1FVeoZOu

To test:
- open AP on desktop
- [x] button tooltip says _Follow up later_?
- now expand QP
- [x] tooltip on the follow-up button says: _Request follow up_
- enter FP
- [x] tooltip on the follow-up button says: Request follow up (this is a different button in the code so you need to check both)
- now click the follow-up button
- [x] blue bar has not more the section name in it?
- click the blue bar
- [x] copy says _x of y section members_?
- [x] is section members clickable with _{section-name} settings_ tooltip only if you are admin?
- [x] does it open the currently section settings for the section selected in cmail (no matter if you are in another section)?
- now open AP on mobile (you can just resize the win and refresh)
- expand a post
- [x] can still complete/create follow-ups with the button at the top
- now enter editing
- click follow-ups button
- click the blue bar
- [x] no underline on _section members_ copy?
- [x] tapping it doesn't do anything?

### DIFF
--- a/scss/partials/_follow_ups_picker.scss
+++ b/scss/partials/_follow_ups_picker.scss
@@ -131,6 +131,18 @@ div.follow-ups-picker {
           line-height: 22px;
           color: $light_navy;
           float: left;
+
+          button.section-memeber-bt {
+            padding: 0;
+            margin: 0;
+            display: inline-block;
+            @include OC_Body_Book();
+            font-size: 14px;
+            line-height: 22px;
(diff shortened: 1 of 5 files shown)


bago2k4:
[Follow-up tooltip copy tweaks](https://trello.com/c/1FVeoZOu/2240-follow-up-tooltip-copy-tweaks)

PR #1041, opened 2019-10-01
Optimize Slack and FileStack image URLs where appropriate

cjsauer:
Trello: https://trello.com/c/7tWvpn1L/2268-optimize-user-avatars-and-filestack-thumbnails

To test:
- Log in with Slack user
- [x] Inspect user avatar. Properly loaded 72px image?
- Upload a custom avatar (i.e. FileStack avatar)
- [x] Inspect your user avatar in the feed. Properly loaded 72px image?
- Create a new post containing a modestly sized image (> 500px)
- [x] Thumbnail image in feed loaded 102px high image?
- Upload a custom org avatar
- [x] Thumbnail image in top-left is pulling from optimized url?

### DIFF
--- a/src/oc/web/images.cljs
+++ b/src/oc/web/images.cljs
@@ -13,7 +13,7 @@
       url
       (let [cur-path    (.getPath uri)
             resize-frag (str "resize=height:" preferred-height)
-            new-path    (str resize-frag "/" cur-path)
+            new-path    (str resize-frag cur-path)
             new-uri     (.setPath uri new-path)]
         (str new-uri)))))
 
--- a/src/oc/web/images.cljs
+++ b/src/oc/web/images.cljs
@@ -1,0 +1,42 @@
+(ns oc.web.images
+  (:require [clojure.string :as cstr])
+  (:import [goog Uri]))
+
+(defn- is-filestack?
+  [uri]
+  (= (.getDomain uri) "cdn.filestackcontent.com"))
+
+(defn optimize-filestack-image-url
+  [url preferred-height]
+  (let [uri (Uri. url)]
+    (if-not (is-filestack? uri)
+      url
+      (let [cur-path    (.getPath uri)
+            resize-frag (str "resize=height:" preferred-height)
+            new-path    (str resize-frag "/" cur-path)
+            new-uri     (.setPath uri new-path)]
+        (str new-uri)))))
+
+(defn- is-slack?
+  [uri]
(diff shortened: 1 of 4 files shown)


src/oc/web/images.cljs
```diff
@@ -0,0 +1,42 @@
...
+            new-path (cstr/replace cur-path re template)
+            new-uri  (.setPath uri new-path)]
+        (str new-uri)))))
+
+(defn optimize-image-url
+  [url preferred-height]
+  (let [uri (Uri. url)]
+    (cond
+      (is-filestack? uri) (optimize-filestack-image-url url preferred-height)
+      (is-slack? uri)     (optimize-slack-image-url url preferred-height)
```
bago2k4: @cjsauer you check `is-filestack?` and `is-slack?` here and then check again inside the specific function to optimize the url. Seems one check is enough.
You can always apply both optimization if you keep the check inside the specific function (`optimize-slack-image-url` and `optimize-filestack-image-url`).
WYT?

cjsauer: @bago2k4 yeah I noticed this too while implementing it. Given that it's essentially just a string comparison, I figured it would be nice for these function to remain useful individually, i.e. it would be nice to be able to call `(optimize-slack-image-url u)` when `u` is _known_ to be a Slack URL.

cjsauer:
> Also we can apply this optimization to the `org-avatar` component that is used to display the org avatar.
> It's not used as much as the `user-avatar` component but still a good thing.

Yeah good call, I'll add that.

src/oc/web/images.cljs
bago2k4: y it makes sense, i was suggesting something like `(-> url (optimize-filestack-image-url preferred-height) (optimize-slack-image-url preferred-height))`.
The check will happen one time inside the function and it will return the initial string if it's not a slack/filestack url.
But no problem, it's only a comparison.

bago2k4:
Seems like we can't optimize Slack images on the fly. They have the size in the URL but that doesn't mean they provide the image at every size. Also the avatar right now have the size appended at the end of the url but that doesn't mean that this is always true.
IMHO we should save a smaller avatar size in our org map so the UI can use it.
I don't think it's worth for the org avatar since it's used in one place only.
I think we should only remove the slack optimization fn but still pass the org avatar through the filestack optimization (if you change the org avatar from our UI it will be saved on filestack).

PR #1044, opened 2019-10-02
Set slack login redirect-uri to expo deep link when on mobile

cjsauer:
Trello: https://trello.com/c/pBS3Fs5W/2257-fix-for-slack-redirect-issue

Related PR: https://github.com/open-company/open-company-mobile/pull/6

To test:
- On mobile staging, try logging in with Slack
- [x] Did you get pushed out to the native browser?
- [x] Are you properly redirected back to the app upon authentication?

Regression:
- [x] Does google login still work as expected on mobile?
- [x] Does email/pass login still work as expected on mobile?

### DIFF
--- a/src/oc/web/actions/user.cljs
+++ b/src/oc/web/actions/user.cljs
@@ -161,11 +161,12 @@
     (api/auth-with-email auth-link email pswd (partial login-with-email-finish email))
     (dis/dispatch! [:login-with-email])))
 
-(defn login-with-slack [auth-url]
+(defn login-with-slack [auth-url & [state-map]]
   (let [auth-url-with-redirect (user-utils/auth-link-with-state
-                                 (:href auth-url)
-                                 {:team-id "open-company-auth"
-                                  :redirect oc-urls/slack-lander-check})]
+                                (:href auth-url)
+                                (or state-map
+                                    {:team-id "open-company-auth"
+                                     :redirect oc-urls/slack-lander-check}))]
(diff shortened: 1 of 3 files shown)


cjsauer:
[Fix for Slack redirect issue](https://trello.com/c/pBS3Fs5W/2257-fix-for-slack-redirect-issue)

src/oc/web/actions/user.cljs
```diff
@@ -161,11 +161,12 @@
...
-(defn login-with-slack [auth-url]
+(defn login-with-slack [auth-url & [state-map]]
   (let [auth-url-with-redirect (user-utils/auth-link-with-state
-                                 (:href auth-url)
-                                 {:team-id "open-company-auth"
-                                  :redirect oc-urls/slack-lander-check})]
+                                (:href auth-url)
+                                (or state-map
+                                    {:team-id "open-company-auth"
+                                     :redirect oc-urls/slack-lander-check}))]
```
bago2k4: hmm @cjsauer not sure this is what we should be doing here.
Before Slack auth was always done using these `team-id` and `redirect` now instead you are going to do it only with the `redirect-origin` passed from the caller.
I think you should be merging the 2 maps here, `state-map` should be the second so it can override the default properties if passed.
WYT?

cjsauer: @bago2k4 yep, resolved 👍

PR #1045, opened 2019-10-04
Fix signup and invite style on mobile

bago2k4:
To test:
- signup from mobile
- [x] style of /sign-up looks good?
- [x] style of /sign-up/profile looks good?
- now invite a new user via email (an email that is not in the system already)
- use the invite link in a new window
- [ ] style of the invite page looks good?
- [x] style of the collect password page looks good?
- [x] style of the collect first/last name page looks good?
- [x] now repeat on desktop both to make sure they didn't regress

### DIFF
--- a/scss/partials/_onboard_wrapper.scss
+++ b/scss/partials/_onboard_wrapper.scss
@@ -165,6 +165,10 @@ div.onboard-email-container {
 div.onboard-lander {
   text-align: center;
 
+  @include mobile() {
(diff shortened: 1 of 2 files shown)


belucid:
@bago2k4 is there meant to be all this white space here?

https://take.ms/2Kv2z

bago2k4:
@belucid sorry about that, fixed the following views, the first one slipped :)
fixed now

PR #1051, opened 2019-10-07
Fix org switch

bago2k4:
Card: https://trello.com/c/43InZpBB

Bug: not sure how to replicate but on our call Stuart was able to switch org w/o actually reloading the app. It was using an internal navigation.
This change should make sure that never happen since it's causing some issues when receiving notification for one or the other (probably because we don't stop and restart the WS connections).

To test:
- go to Carrot AP on staging
- open Dev tools on Network tab
- make sure Preserve log is unchecked
- click on Joy restaurant
- [x] make sure all the old network requests are cleaned out
- [x] make sure the first request is a GET to /the-carrot-stand/all-posts
- [x] repeat for another org

### DIFF
--- a/src/oc/web/components/ui/org_avatar.cljs
+++ b/src/oc/web/components/ui/org_avatar.cljs
@@ -3,11 +3,7 @@
             [org.martinklepsch.derivatives :as drv]
             [oc.web.urls :as oc-urls]
             [oc.web.images :as img]
-            [oc.web.router :as router]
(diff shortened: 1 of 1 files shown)


bago2k4:
[Org switching is too fast! (WS connections are from old org)](https://trello.com/c/43InZpBB/2287-org-switching-is-too-fast-ws-connections-are-from-old-org)

PR #1060, opened 2019-10-30
Working prototype of stream refresh on window focus

cjsauer:
Trello: https://trello.com/c/8aUd0R9R/2328-ensure-post-streams-are-always-fresh-when-returning-to-carrot

To test:

- Comment out the body of the [interaction client's `start-router!` code](https://github.com/open-company/open-company-web/blob/bb3d814f0469fdcd4befed43ae21e25b4cb34568/src/oc/web/ws/interaction_client.cljs#L161-L164)
- Visit AP
- Put this browser window in the background
- In a different browser session, create a new post
- [x] Ensure you do NOT see the post (you commented out the WS code, right?)
- Re-focus the original browser window
- [x] Does the new post load?
- Change sections and repeat this test
- [x] Works in AP, follow-ups, and normal sections?


bago2k4:
Seems good, i think we should improve this and check the WS connections when the focus event happens. If they are disconnected we need to reconnect them.

PR #1063, opened 2019-11-04
Hide thumbnails from mobile FoC

bago2k4:
Card: https://trello.com/c/nwg6rvxo

To test:
- add a few posts, some with an image, other with a video etc..
- [ ] check that you see the thumbnail on desktop
- [ ] check that you don't see the thumbnails on mobile
- check that everything is ok in expanded post page


bago2k4:
[Mobile only: Remove thumbnails from FoC](https://trello.com/c/nwg6rvxo/2334-mobile-only-remove-thumbnails-from-foc)

PR #1065, opened 2019-11-05
Fix time format

bago2k4:
Card: https://trello.com/c/kQeLIQIf

To test:
- add a new post
- [ ] check in FoC that you see the publish time and not a relative time (1:12 AM instead of 2h)
- [ ] make sure you don't see a leading zero before in the hour
- [ ] make sure AM/PM are uppercase
- [ ] from REPL make sure that `oc.web.lib.utils/foc-date-time` return the time without the initial zero for today's hours with a single digit and that 2 digits are returned for 11 and 12
- [ ] also make sure AM/PM are always upper case


bago2k4:
[Fix FoC time format](https://trello.com/c/kQeLIQIf/2338-fix-foc-time-format)

PR #1067, opened 2019-11-05
Compose tweaks

bago2k4:
Card: https://trello.com/c/SHwG2y8J

To test:
- on desktop go to AP (or another stream)
- expand QP
- [x] do you see the new design?
- pick a section
- leave subject empty
- add some body
- [x] do you see the empty subject tootlip when you hover the post button? (make sure it says subject and not title)
- now add a title
- [x] publish the post, did it work?
- now add a post from FP
- leave subject empty
- [x] do you see the tooltip when you hover the post button?
- add a title
- type something in the body
- [x] do you see the "Add a quick summary" button appear when you type 255 chars?
- click it
- [x] does the abstract appear?
- type some text in the abstract and overflow the maximum length
- [x] do you see the tooltip on the post button?
- remove some text from the abstract but leave some text in it
- now click the X in the top right corner of the abstract
- [x] does the abstract disappear?
- [x] do you see the post being saved?
- [x] do you see the "Add a quick summary" button appear again (only if the body is still at least 255 chars)?
- now click "Add a quick summary" again
- [x] does the abstract field appear again?
- [x] does it have the old content still in it?
- publish the post
- [x] do you see the abstract being kept?
- now edit the just published post
- [x] do you see the abstract field visible right away?
- remove some body so it's very short (less than 255 chars)
- [x] do you NOT see the abstract field disappear?
- Save the post
- play some more with the post fields to make sure the abstract field appear and disappear when it should
- switch to mobile
- [x] do you see the abstract field when you edit a post that has one?
- [x] do you NOT see the abstract field or the "Add a quick summary" button when you add a new post or edit a post w/o one?
-


bago2k4:
[Compose tweaks](https://trello.com/c/SHwG2y8J/2329-compose-tweaks)

PR #1071, opened 2019-11-20
ClojureScript upgrade

bago2k4:
Cards:
- https://trello.com/c/4ck8AkOv
- https://trello.com/c/S6dzopQN

Review with: ~~https://github.com/open-company/open-company-lib/pull/77~~

Upgrade to ClojureScript 1.10.597 and all other outdated dependencies.

I have also updated React to 16.11.0 which is the latest. The only dependency I didn't update is FileStack lib since that needs some code changes for how we handle files.

To test check all the main feature of the client on staging.

NB: I changed Clojure version from 1.10.1-RC1 to 1.10.1. The new prod server is already running on this Clojure version, but the current beta is still on 1.9. Not sure if this can be a problem. On staging is running just fine. If you think it's a problem we should wait and merge this when we switch to the new prod server.


bago2k4:
[Web deps upgrade](https://trello.com/c/4ck8AkOv/2365-web-deps-upgrade)

bago2k4:
[Upgrade React to v16.5+ to use profiler](https://trello.com/c/S6dzopQN/2263-upgrade-react-to-v165-to-use-profiler)

bago2k4:
@belucid removed Ready for review from this since I just noticed some warnings on staging about `componentWillMount`, `componentDidMount` and `componentDidReceiveProps` methods, of which we are making large use.

They will be deprecated in React 17 so they triggers some warnings. We can keep it and live with the warnings until we change how our components work or we avoid upgrading to React 16.11 (not sure right now what's the max version we can use).

belucid:
@bago2k4 we'll live with the warnings for now. Important reminder of the tech. debt there to move off those.

PR #1088, opened 2019-12-16
Mobile new comment label fix

bago2k4:
Card: https://trello.com/c/W3cCNPaC

To test:
- change your user name to a very long name
- with another user add a few comments to a post
- open the post on mobile
- do you see the user name truncated so the timestamp and NEW can fit on a single line?
- repeat on desktop


bago2k4:
[Mobile new comment label fix](https://trello.com/c/W3cCNPaC/2388-mobile-new-comment-label-fix)

PR #1092, opened 2019-12-19
Autosave sync

bago2k4:
Card: https://trello.com/c/70GuYBKj

Related PRs:
- [Storage](https://github.com/open-company/open-company-storage/pull/228)
- ~~[Change](https://github.com/open-company/open-company-change/pull/24)~~
- ~~[Lib](https://github.com/open-company/open-company-lib/pull/79)~~

To test concurrent edits:
- open 2 sessions of the same user
- with session 1 create a new draft
- with session 2 go to drafts and start editing the same post
- make the edit you make in one session are shown in the other session

To test client race condition on post publish:
- open QP
- type some body and headline quickly
- click Post as soon as you see Saving...
- [x] the post was published?
- [x] no twin post (same headline and body but different uuid) in drafts?

NB: this could have weird outcome in case of users trying to do collaborative editing, for example one user could override the edits of the other. The last edit wins (updated-at is used to compare).


belucid:
@bago2k4 Problem scenario:

Follow the test steps above for concurrent drafts open in 2 tabs.
Delete the draft in 1 tab.
Other tab loses its "Post to" value and upon save errors.

bago2k4:
@belucid fixed

bago2k4:
Are we sure losing the post on every window is what we want here?
If you delete the post from one session it will be closed on all the others too, it means the post is lost.
The delete action has a confirmation dialog so it's pretty safe. Just thinking out loud..

belucid:
> Are we sure losing the post on every window is what we want here?
> If you delete the post from one session it will be closed on all the others too, it means the post is lost.
> The delete action has a confirmation dialog so it's pretty safe. Just thinking out loud..

Yeah, given the confirmation dialog we're sure that's what we want.

belucid:
@bago2k4 can't delete a draft in this branch, it just 404's

bago2k4:
@belucid this is ready again but you should review this urgent PR first https://github.com/open-company/open-company-storage/pull/232

PR #1094, opened 2020-01-15
Update marketing site

bago2k4:
### Review after #1106 and not before we have updates Stripe pricing

Card: https://trello.com/c/gJZTGN62


belucid:
waiting on stripe for free change

PR #1097, opened 2020-01-20
Invite via Slack tweaks

bago2k4:
From our testing [feedback card](https://trello.com/c/W3zEkMYC):
> The check for "Invite via Slack" link being enabled in the panel shouldn't depend on them being an admin, should be enabled if there is a Slack org associated with the team. Just don't show the bottom panel "Enable the Carrot bot for Slack" if they aren't an admin.

To test:
- signup for an email org
- invite a contributor
- both users see Invite via Slack disabled?
- signup via Slack (do not add the bot yet)
- invite a contributor via email
- admin sees invite via Slack enabled?
- contrib sees Invite via Slack disabled?
- invite a contributor via Slack (do not add the bot, use the Slack link)
- new contrib sees Invite via Slack enabled
- with the admin click Invite via Slack
- does it gets the add bot button at the bottom?
- click it
- after adding the bot the admin sees invite via Slack with users picker too?
- both contributors can enter Invite via Slack and can invite with link and with user's picker?


belucid:
@bago2k4 for both admin and contrib, the "Invite via Slack" looks disabled. But for both it's a link. For an admin the link works. For contrib the link gives a `TypeError: null is not a valid URL.`.

bago2k4:
Removed the Invite via Slack button completely if it's disabled and the user is not an admin.

Only admins can add the bot, before we were sending all users through auth step first and then the bot step.

Now we do directly the second.

We could still make go through the first auth step so they associated a Slack org to the team and then we can let them copy the Slack invite.

Not sure it's worth

PR #1103, opened 2020-01-22
Comments fixes

bago2k4:
This should fix a couple of small bugs around comments:
- add the first comment on a post: see the summary still says Add comment
- add comment field is white after a comment is added, should show the placeholder
- publish date and comments dates have no tooltip showing the full date and the last edit

To test:
- add the first comment to a post
- summary changed to 1 comment?
- add some more root comments and replies and check the summary is updated as expected
- make sure the add comment field has always the placeholder visible initially
- check the tooltip shows up when you hover the post publish date in expanded view
- check the tooltip shows up when you hover a comment date
- copy public share url
- open in incognito window
- check post has a tooltip on the date
- copy digest url (with id-token)
- open in incognito window
- check comments summary is right
- check post and comments have a tooltip on the date


belucid:
@bago2k4 this one has conflicts.

belucid:
@bago2k4 on this branch if you leave a root comment, and then immediately add a reply to it, the reply button never enables.

Tested this on prod and it's not an issue there.

bago2k4:
@belucid good catch, fixed

belucid:
hmm I don't understand the last testing instructions @bago2k4 

we don't show any comment summary in public share it seems

PR #1107, opened 2020-01-23
Edit attachments fixes

bago2k4:
Bug: if you add an attachments while editing a draft/post (full post editing)that has some body and/or abstract, the file will be added at the bottom of the page, sometimes it will be completely out of sight.

This moves the attachments list/update state up, between the abstract and the body.

To test:
- add a new post
- enter full post edit
- add some body until you get the abstract prompt
- add some text in the abstract field
- add a file (D&D or with the 📎 button)
- upload percentage is shown above the body?
- when upload finishes the file is attached in the same place?
- publish the post
- check mobile edit for regressions


belucid:
@bago2k4 conflicts here.

bago2k4:
@belucid conflicts addressed

PR #1109, opened 2020-01-27
Fix FoC menu

bago2k4:
When in FoC if you click on the ... menu and then move the cursor out of the post tile, the menu gets cut off.

To test:
- on desktop with expanded FoC
- open the menu
- move cursor out of the post tile
- menu is hidden?
- the FoC buttons (bookmark, follow, share) have the same space from right and top border of the tile?
- on desktop with collapsed FoC
- open the menu
- move cursor out of the post tile
- menu is hidden?
- on mobile
- check you can still open the menu with swipe or long press


bago2k4:
[Post deploy tweaks](https://trello.com/c/9SYcQJGI/2423-post-deploy-tweaks)

PR #1112, opened 2020-01-30
Dark mode tweaks

bago2k4:
Card: https://trello.com/c/aTJ0wGQt

To test:
- check all the menu panels with a long window
- everything fine?
- check all the menu panels with a very short window (not mobile)
- everything fine? you can scroll the content in all panels?
- now check the theme panel
- check the system settings box
- refresh and make sure the status is preserved
- select dark
- refresh and make sure the status is preserved
- select light
- refresh and make sure the status is preserved
- check it on mobile too

Test preventing dark mode on certain routes:
- now change your Carrot setting to dark mode
- visit /login (incognito if needed)
- do you never see dark mode (for example not even when there is the shaking carrot)?
- visit /reset
- same as above?
- visit /404
- same as above?
- check randomly some of [these routes](https://github.com/open-company/open-company-web/pull/1112/files#diff-c7ec236122d2d25629e0cd6a0a32f299R344-R357)
- same as above?

NB: the foreground event is a bit hard to test, it attaches a listener to the document object for the `visibilityChange` event. When that's triggered, if the window is visible, it checks if user has the theme set to auto, if so it sets the theme to the settings value.


bago2k4:
[Dark mode: foreground event and Theme](https://trello.com/c/aTJ0wGQt/2409-dark-mode-foreground-event-and-theme)

PR #1116, opened 2020-01-31
Fix WRT count

bago2k4:
Card: https://trello.com/c/5sNHBWCp

Bug:
- open Unread
- nav to AP,
- check Viewers is show for some posts (those that were not visible in Unread)
- nav to Bookmarks
- back to AP
- now you can see the counts

Problem is that the stream-item component doesn't get updated when the read counts are loaded.

To test:
- repeat the steps above
- confirm the bug is solved
- check switch to collapsed item for regressions
- check mobile for regressions
- merge


bago2k4:
[Viewers count issue](https://trello.com/c/5sNHBWCp/2436-viewers-count-issue)

PR #1117, opened 2020-02-03
Fix word wrapping

bago2k4:
Found out that W3C recommends the use of `overflow-wrap` for english aphabet languages, and the use of `word-wrap` for non-english ones.

To test:
- add a post with a very long single word in title, post and abstract
- check it's displayed fine in: stream, collapsed stream, expanded post
- add a comment with a very long single word
- check that the comment is displayed without issues
- copy the publish share url of the post
- open in incognito
- check is displayed fine there too

### DIFF
--- a/scss/mixins/_activity.scss
+++ b/scss/mixins/_activity.scss
@@ -134,6 +134,8 @@ $h2_top_margin: 24;
   font-size: $font_size;
   color: var(--color);
   line-height: $line_height;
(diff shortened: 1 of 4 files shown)


bago2k4:
[Weird spacing in Abstract once posted](https://trello.com/c/XDcKeHSE/2437-weird-spacing-in-abstract-once-posted)

PR #1119, opened 2020-02-04
Fix commenting on ID token view

bago2k4:
Card: https://trello.com/c/VkdkVn9b

To test:
- get a digest link with id-token
- open in incognito
- focus on comment field
- add some words
- click the X to cancel out
- did it work?
- now add a comment and post it
- add a reply to that comment with a mention
- now edit one of the comments you just added

### DIFF
--- a/scss/partials/_secure_activity.scss
+++ b/scss/partials/_secure_activity.scss
@@ -297,7 +297,7 @@ div.secure-activity-container {
     }
 
     button.secure-activity-footer-bt {
-      height: 18px;
+      height: 72px;
       @include OC_Body_Book();
       font-size: 14px;
       color: $carrot_green;
--- a/src/oc/web/actions/activity.cljs
+++ b/src/oc/web/actions/activity.cljs
@@ -584,7 +584,7 @@
 (defn org-data-from-secure-activity [secure-activity-data]
   (let [old-org-data (dis/org-data)]
     (-> secure-activity-data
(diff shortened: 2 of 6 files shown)


belucid:
@bago2k4 in the test sequence, after you X out of the comment and then click in again, the caret cursor is poking out of the top of the comment box (on FF anyway).

Can we check if comment is blank before showing them the close confirmation dialog? Right now we confirm them losing nothing.

bago2k4:
@belucid fixed the issue with FF. Moved the fix also to the body editor of Cmail since it was affected too.

About the check on blank: checking if the field is empty is not very quick, since that's an HTML field i need to check if there is text but also pictures or videos or other "solid" elements. That means interpreting the html code and parse it, I think we can do that w/o impacting performance but I would like to do it on its own.
WYT if we add a card for it and prioritize?

Also if/when we decide to add other rich content elements we need to remember to expand this check.
Say we now check if body has: no text, no `iframe` elements (this is what we use for videos now), no `img` or `picture` elements.
One day we decide to add `video` tags, we need to remember to add that to the check.

belucid:
@bago2k4 no problem then on the dialog, we'll get to it some other time maybe. Not important for now and no need to track it in a card.

bago2k4:
Forgot about this: Stuart was asking about this too for the body field. He wanted to see the placeholder appearing again when completely deleting the content of the entry body or a comment.

I said it wasn't quick so we decided to wait, so working on this could fix that too.

PR #1136, opened 2020-03-02
Compose button

bago2k4:
Card: https://trello.com/c/E809Ds0X

To test:
- open AP
- plus button in top right is grey?
- scroll down to partially cover QP
- plus button turns green?
- click it
- close Cmail
- switch to a section
- expand QP
- current section is the default one?
- close QP
- add a new section
- expand QP
- new section is selected?
- switch to another section
- check it's selected in QP
- delete the section
- check the deleted section is no longer selected in QP
- check the deleted section is no longer in the sections list of QP
- type something in QP
- click the + button in top right
- still editing the same post but in fullscreen?
- play a bit with QP and FP and sections switching


bago2k4:
[If QP not visible, new Compose +](https://trello.com/c/E809Ds0X/2469-if-qp-not-visible-new-compose)

belucid:
Feedback from the call.

belucid:
@bago2k4 CI failing here

bago2k4:
@belucid CI is fixed, sorry for the delay

belucid:
@bago2k4 strange issue, in Cmail, normally all the tooltips of the top bar icons (close, delete, emoji, ...) drop down (edited) but if you scroll down enough for the + button to turn green, then click it, close and attach icon tooltips open upward instead and so they can't be seen.

bago2k4:
@belucid fixed.
Cmail fullscreen locks the page scroll so you think you are at the top of the page (ie: `scrollTop = 0`) but the page is actually still at the same scroll position.

This was confusing bootstrap that was calculating the space above the buttons to check if the tooltip could be shown above.

This fix makes sure the page is scrolled to the top before locking the scroll so bootstrap can do its job w/o problems

PR #1141, opened 2020-03-13
Onboard fields fix

bago2k4:
Fix: the onboard form sometimes is updated after the user has started typing. That happens because the user data are loaded with some delay and once that happens the values from the server are inserted in the fields and the name one is focused.

To test:
- onboard via Slack
- once you land on team setup lander (after granting Slack permissions) start typing ASAP in the Why carrot field
- did it NOT move the focus to the name field after you started typing?
- check that inviting via team email link still works and is not affected by the above issue (start typing first name asap)
- check that inviting via email still works and is not affected by the above issue (start typing first name asap and check your input is not overwritten)


bago2k4:
[Code blocks (no syntax highlighting)](https://trello.com/c/6MRWcY4W/2252-code-blocks-no-syntax-highlighting)

belucid:
@bago2k4 I'm not seeing any change in behavior on this one.

I still get focus going back to name after starting typing in Why. Also Why isn't the active field to start after Slack login, though it should be.

bago2k4:
I removed the delay in focus now, before it was waiting 2.5 seconds before focusing on the first field of the view after the component was mounted.
Now it focus on it right away so you shouldn't be able to type anything before.

I wouldn't change the focus, not sure how much it can take since we don't wait for all the data to be loaded when this view is shown and to know if user is using Slack or email i need more data.

Hope it makes sense

PR #1143, opened 2020-03-14
Fix Sentry browser

bago2k4:
NB: this is urgent only because right now in Sentry we get only the errors and the stack traces, no info about OS, browser and deploy release version.

Review with:
- [Infra](https://github.com/belucid/open-company-infrastructure/pull/130)

Sometimes Sentry doesn't add the OS and browser info.
Also `oc.web.lib.sentry/capture-error-with-context!` and `oc.web.lib.sentry/capture-error-with-message` are broken since they don't catch the error they create, so we get a report but w/o all the needed details and also it's not handled.

To test:
- open staging
- in console run `oc.web.lib.sentry.test_sentry()`
- do you see a message in Sentry? (Message from clojure)
- do you see also an error? (`ReferenceError: errorThrowingCode is not defined`)
- do you see release, user, OS and browser info on both?
- now run `oc.web.lib.sentry.capture_error_with_message("My first error", "My error message");`
- check the error contains release, user, OS and browser info


belucid:
@bago2k4 I don't get anything in Sentry on staging after `oc.web.lib.sentry.capture_error_with_message("My first error", "My error message");`

bago2k4:
@belucid good catch. I fixed it.

belucid:
@bago2k4 I still don't get anything in Sentry on staging after oc.web.lib.sentry.capture_error_with_message("My first error", "My error message");

bago2k4:
@belucid now it should work. I had to add a PR for infra since we were limiting the errors to those that happen on scripts loaded under `staging.carrot.io` but our js code is loaded form the CDN.

You can find the link in the description

belucid:
@bago2k4 I get a `Sentry Logger [Error]: Error while sending event: [object Response]` when running `oc.web.lib.sentry.capture_error_with_message("My first error", "My error message");` on staging now.

bago2k4:
@belucid fixed as per the discussion on Slack

PR #1146, opened 2020-03-17
Redirect to AP on empty inbox

bago2k4:
Card: https://trello.com/c/zPYZ5VyY

To test:
- load inbox when there is at least 1 post
- dismiss that post
- refresh inbox
- landed in AP?
- now nav to inbox using the navigation sidebar
- did it load inbox?
- double click on inbox label in the navigation sidebar
- still in inbox?
- nav around the app and check you are never redirected to AP when inbox is empty except on a page refresh


bago2k4:
[Redirect to Recent if nothing in Unread](https://trello.com/c/zPYZ5VyY/2482-redirect-to-recent-if-nothing-in-unread)

PR #1137, opened 2020-03-03
Code blocks

bago2k4:
Card: https://trello.com/c/6MRWcY4W

Related PR:
- [Interaction](https://github.com/open-company/open-company-interaction/pull/43)
- [Lib](https://github.com/open-company/open-company-lib/pull/80)

To test:
- try adding a code block with multiple lines using the picker button
- try adding a code block with multiple lines using the ``` starter
- try adding an inline code using the format button
- try adding an inline code using the ` starter/closer
- make sure you can't format text inside code blocks (both type)
- copy some text that include a `PRE` element from an HTML page and paste it in our editor
- do you see the PRE elements transformed into code?
- do you NOT see text formatted with bold or italic inside etc?
- repeat the copy/paste with a `CODE` element
- test block codes in comments
- make sure digest looks good if there is a post with code blocks
- make sure Slack/email shares looks good if there is a post with code blocks
- make sure notifications (in-app, email, slack) looks good if there are code blocks
- make sure code blocks look good in FoC
- make sure code blocks look good in dark mode

PR #1148, opened 2020-03-17
Add h1 and highlighter to text formatters.

bago2k4:
Card: https://trello.com/c/s7w5BfC1

NB: interaction service will remove highlight from the body (not the text, just the background color) since it doesn't accept mark elements right now. It changes with [code-blocks PR](https://github.com/open-company/open-company-web/pull/1137)

To test:
- test H1 formatter
- test H2 formatter
- test Highlighter formatter
- use a mix of them
- make sure you can always type outside of a highlighter portion of text


bago2k4:
[Format (H2, highlight)](https://trello.com/c/s7w5BfC1/2449-format-h2-highlight)

belucid:
@bago2k4 H2 icon is H1

bago2k4:
@belucid  i think the icon bug you found on format PR is just cache. Before we had H1 but we were actually using an h2 tag in code, so the icon was named one way but represented the other thing. Try forcing refresh and it should be clear. I double checked on staging and i see the correct icons, at least in the ME context menu

Screenshot 2020-03-24 at 3 53 35 PM

PR #1157, opened 2020-04-13
Fix disappearing placeholder

bago2k4:
Card: https://trello.com/c/2ISZFUiS

Not really "urgent" since it's not blocking or causing big issues. It's just annoying that drafts are automatically auto-saved when you expand QP (meaning you are going to have a lot more drafts than you would expect).

NB: this is causing also another issue: when you click on QP to start writing it saves right away. It happens because the initial body is empty (ie: "") but as soon as you focus it ME initialize it with an empty paragraph which is seen as a change action.
Initializing the editor with an empty paragraph from the beginning prevents this behavior.

To test QP:
- open app
- do you see the placeholder on QP?
- do you see the MediaPicker buttons?
- click on QP
- did it expand?
- placeholder still visible?
- did it NOT autosave?
- type something
- did it autosave?
- placeholder went away?
- test a few things like: board switching, add media, fullscreen switching/editing, publish etc

To test add comment field:
- open app
- click on a post
- is add comment visible but collapsed?
- do you see the placeholder?
- click on it
- placeholder still visible?
- type something
- placeholder went away
- type some more
- click back arrow
- re-open the same post
- do you see your typed comment still in the field?
- is the add comment expanded (post button is visible)?
- post a comment
- repeat the cache test with a reply field
- do some more testing around comments: comment/reply with media, post, edit, delete etc...

### DIFF
--- a/scss/partials/_add_comment.scss
+++ b/scss/partials/_add_comment.scss
@@ -8,6 +8,10 @@ div.add-comment-box-container {
       cursor: text;
       padding: 12px 16px;
 
+      p:first-child {
+        display: none;
+      }
+
       @include mobile() {
         padding: 11px 14px;
       }
--- a/scss/partials/_cmail.scss
+++ b/scss/partials/_cmail.scss
@@ -91,6 +91,10 @@ div.cmail-outer {
               div.rich-body-editor {
                 margin-top: 0;
                 min-height: 24px;
+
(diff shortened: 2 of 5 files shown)


bago2k4:
[Lost placeholder](https://trello.com/c/2ISZFUiS/2516-lost-placeholder)

PR #1166, opened 2020-08-30
Leiningen and shadow-cljs migration

bago2k4:
I moved the web app over to shadow-cljs: goodbye boot, but more important, goodbye cljsjs, and welcome lein/shadow-cljs/npm.

At this point only the web app works, you can start it with: `npm run dev:watch` and then visit http://localhost:3559.

Since only the web app is working there is no static site or electron app configuration.

I added a lein configuration file to create the static HTML files from hiccup since we need to read some environment variable while building. I think i'll have a lein alias at the end starts everything so we don't have to run more tasks to work on the web app.
Instead of `lein build` -> `npm run dev:watch`, I should be able to start the npm task directly from a lein alias once the site has been built without errors.

Next steps to finish up this migration are:
- finish lein configuration for the static website and the app shell build
- make sure production build still works
- change deploy playbook in infrastructure
- fix electron web app
- fix expo web app (mobile)


bago2k4:
Closed since the branch was changed to `shadow-cljs` since this one had issues with the moved files.

Since file were not moved directly from git it was unable to trace back changes on the old code base to here, that means every change would need a manual port.

With the new branch we should be able to just merge the old code base into the new one and fix minor conflicts if preset.

